PR #3, opened 2018-08-18
[WIP] Implement interface to provide the resolver an initial state

### LINKED ISSUE #1: Support single package upgrade semantic
The resolver should be able to accept an already resolved graph, and start resolving from there (instead of a new graph). Upgrade semantic can probably be done on the callee side: Remove a vertex from an existing graph [*], and send it to the resolver with the requirements (including the package to be upgraded) to upgrade. I don’t have a proof, but this should work.

[*]: The vertex’s children should not be removed (although edges to it obviously need to be severed). The resolver will probably need to gain a cleanup phase to remove orphan vertices (vertices without any parents) from the graph after resolution is done.

uranusjr:
This would allow add, remove, and upgrade semantics. Fix #1.

* [x] Enable passing in a state object.
* [x] Public state interface.
* [ ] Review internals to be resillient of slightly out-of-sync mapping and graph
* [ ] Resolver should clean up the graph to purge dead subtrees, and maping to remove unneeded pins after resolution (or maybe after each round?).
* [ ] `add` semantic
* [ ] `remove` semantic
* [ ] `upgrade` semantic

### DIFF
--- a/news/1.feature.rst
+++ b/news/1.feature.rst
@@ -1,0 +1,1 @@
+Support resolving based on an existing state, instead of always from ground zero. This allows the user to perform a differential resolution that only makes minimal changes to a given dataset.
--- a/news/3.behavior.rst
+++ b/news/3.behavior.rst
@@ -1,0 +1,1 @@
+The `State` class is not part of the public interface, and is a “real” object instead of a tuple subclass.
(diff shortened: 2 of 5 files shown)


uranusjr:
I’m starting to think this might not be the right approach for our particular purpose. We can just bump the already locked version in Passa to get it locked again. Closing this for now to try working on Passa first.

techalchemy:
Agree. This should just resolve

PR #9, opened 2019-02-18
Add AbstractResolver class and introduce (and reorganize) 'abstract' subpackage

pradyunsg:
@techalchemy This is what I was suggesting -- adding an 'AbstractResolver'.

This would basically make resolvelib's resolver-like interface implemented separately (and provide different semantics like PubGrub-based resolution or Molinillo-based resolution).

### DIFF
--- a/src/resolvelib/__init__.py
+++ b/src/resolvelib/__init__.py
@@ -1,6 +1,6 @@
 __all__ = [
     '__version__',
-    'AbstractProvider', 'BaseReporter', 'Resolver',
+    'AbstractProvider', 'AbstractResolver', 'BaseReporter', 'Resolver',
     'NoVersionsAvailable', 'RequirementsConflicted',
     'ResolutionError', 'ResolutionImpossible', 'ResolutionTooDeep',
 ]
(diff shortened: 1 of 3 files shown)


pradyunsg:
I also moved all the abstract classes into an 'abstract' subpackage, mostly because IMO it is cleaner to explicitly separate abstract classes from actual implementation code.

pradyunsg:
I didn't rename BaseReporter for maintaining backwards compatibility and also because I don't think it's a problem that it is in the `abstract` subpackage and not called `Abstract...`.

uranusjr:
+1 for base classes, but I don’t think abstract classes need a submodule; I’d prefer to put `AbstractResolver` in `providers.py` and have everything else stays as-is.

`BaseReporter` is named base since it is not abstract; it is entirely valid to instantiate an instance (and get no reporting at all). Maybe I should have named it *Empty* instead…

pradyunsg:
Ok to go?

pradyunsg:
> Maybe I should have named it _Empty_ instead…

I'd go with `NoOpReporter`

PR #15, opened 2020-01-31
Add an example provider for wheels from PyPI

pfmoore:


pfmoore:
@uranusjr As discussed, I've packaged up my demo PyPI provider as an example for the sources. Still needs some better comments/documentation, but it's functional as it stands.

uranusjr:
I read through the implementation (didn’t actually run it) and everything checks out for me. It’s refreshing to see how little code it needs if we’re allowed to get rid of all the historic baggage 😄

uranusjr:
I’ll just merge this and give you write access so you can work on it directly.

pfmoore:
Nice! Thanks. Yeah, this turned out to be a good example of how simple pure standards-based code can be, as well :-)

PR #17, opened 2020-02-01
Functional tests from Swift Package Manager

uranusjr:
I *think* I’ve got most things right, but 3/4 of the resolution results are slightly different (each with one discrepancy). Not sure if this is a functional difference between implementations; will need to look into those.

```python
# Kitura
{'https://github.com/IBM-Swift/BlueSocket.git': '0.12.0'} != {'https://github.com/IBM-Swift/BlueSocket.git': '0.12.4'}
{'https://github.com/IBM-Swift/CCurl.git': '0.3.7'} != {'https://github.com/IBM-Swift/CCurl.git': '0.2.3'}

# PerfectHTTPServer
{'https://github.com/PerfectlySoft/PerfectLib.git': '2.0.0'} != {'https://github.com/PerfectlySoft/PerfectLib.git': '2.0.4'}
{'https://github.com/PerfectlySoft/Perfect-Net.git': '2.0.0'} != {'https://github.com/PerfectlySoft/Perfect-Net.git': '2.0.5'}

# SourceKitten
# Passes \o/

# ZewoHTTPServer
{'https://github.com/Zewo/CLibvenice.git': '0.15.0'} != {'https://github.com/Zewo/CLibvenice.git': '0.14.0'}
```

Assuming both are correct, neither resolver is “superior” IMO since each got newer packages in some cases.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,8 @@
+import pytest
+
+from resolvelib.reporters import BaseReporter
+
+
+@pytest.fixture(scope="session")
+def base_reporter():
+    return BaseReporter()
--- a/tests/functional/swift-package-manager/inputs/LICENSE
+++ b/tests/functional/swift-package-manager/inputs/LICENSE
@@ -1,0 +1,211 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+    TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+    1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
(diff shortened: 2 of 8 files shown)


uranusjr:
For `ZewoHTTPServer`, all requests to `CLibvenice` ask for `{'0.14.0', '0.15.0'}`. `CLibvenice` has no dependencies. Therefore both `0.14.0` and `0.15.0` are correct, and ResolveLib picks the higher one (as implemented in the provider).

(Note: Four candidates ask for `CLibvenice`: `Venice 0.14.0`, `File 0.14.0`, `TCP 0.14.0`, and `IP 0.14.0`.)

The reason Swift Package Manager chooses the lower one may be the [Minimal Version Selection](https://research.swtch.com/vgo-mvs) principal, which prefers lower versions to higher if both satisfy the constraints.

This should theoretically be implementable by modifying `find_matches()` to prefer candidates with lower versions (i.e. put them last instead of first). But interestingly this causes the resolver to throw `ResolutionImpossible`. It is clear we are missing some backtrack possibilities.

uranusjr:
Merging since I am implementing the tests correctly. The fact they fail is a good thing. The causes should be tracked by subsequent issues.

PR #19, opened 2020-02-01
Backtrack on exhausted candidates

### LINKED ISSUE #18: Backtrack when no candidates are found
See https://github.com/sarugaku/resolvelib/pull/17#issuecomment-581033508

The `ResolutionImpossible` was thrown here, during when we go through the candidates to find one that satisfies all existing constraints. Currently we give up when all candidates fail. We should instead backtrack (pop the state stack and remove the last pinned candidate) and keep trying.

https://github.com/sarugaku/resolvelib/blob/49214e76e212f2218de62cf05c15ebc0c41dbc56/src/resolvelib/resolvers.py#L207-L216

uranusjr:
Fix #18.

This is done by repeatedly restoring old states and fix them one by one, attempting to apply our knowledge on dead criteria:

* Mark candidates pointing to the dead criteria as incompatibilities.
* Remove the dead criterion.

If any of the state can handle these fixes (without causing more incompatibilities), use it as our resume point, and start forward from there.

I was always wondering why all the resolvers need to store a `known_incompatibilities` property. Now I know (I think).

### DIFF
--- a/news/18.bugfix.rst
+++ b/news/18.bugfix.rst
@@ -1,0 +1,4 @@
+Make resolver backtrack when none of the candidates requested for a package
+are able to resolve due to them requesting unworkable requirements, or a
+package has no candidates at all. Previously the resolver would give up on the
+spot.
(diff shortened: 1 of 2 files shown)


uranusjr:
Oh and in case anyone wonders, all test cases from Swift Package Manager are now passing :)

Issue #4, opened 2018-08-19
Clean up graph after resolution


uranusjr:
Resolver should clean up the graph to purge dead subtrees, and maping to remove unneeded pins after resolution (or maybe after each round?).

PR #21, opened 2020-02-02
Test cases from CocoaPods

uranusjr:
Gosh now I need to implement Ruby’s version specifiers.

https://guides.rubygems.org/patterns/#declaring-dependencies


pradyunsg:
An option would be to "translate" them into Python equivalents. :)

uranusjr:
`swapping_children_with_successors` is detecting superfluous graph items due to #4. I guess it is a valuable thing to do.

uranusjr:
Merging first and fix the xfails later.

PR #23, opened 2020-02-02
Add Travis configs

uranusjr:
Also updates package metadata to meet modern standards. Still based on Setuptools because I don’t want to make too many changes at once.


uranusjr:
How am I supposed to enable Travis…

uranusjr:
Yay

PR #26, opened 2020-02-04
Add test vector on alternative pin ordering

uranusjr:
New failures, yay!

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -6,3 +6,16 @@ from resolvelib.reporters import BaseReporter
 @pytest.fixture(scope="session")
 def base_reporter():
     return BaseReporter()
+
+
+def constrained_first(resolution, candidates, information):
+    return len(candidates)
+
+
+def loose_first(resolution, candidates, information):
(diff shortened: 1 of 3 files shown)


uranusjr:
Uh, I think there are fundamental problems in the resolving process that needs some rewrite. In short, the resolver is not properly backtracking dependency “contributions” when it removes a previous pin that is no longer compatible due to incoming new requirements.

I am walking away from this for a while since I’ve worked too much and feel a little tried about thinking this. Let’s focus on improving other aspects first, we can always swap out the internals afterwards.

pradyunsg:
> feel a little tired about thinking this.

/me relates strongly.

> In short, the resolver is not properly backtracking dependency “contributions” when it removes a previous pin that is no longer compatible due to incoming new requirements.

IIUC, the issue is with https://github.com/sarugaku/resolvelib/blob/master/src/resolvelib/resolvers.py#L86?

pradyunsg:
Closing this out for now -- we can come back to this later, if we find that we need to.

Issue #27, opened 2020-02-05
Adding Python-esque test cases


pradyunsg:
One of the things that can be found in zazo's issue tracker, is a bunch of Python-specific failure cases for the current pip resolver.

It might be useful to add test cases for those situations. Another case that would be really handy would be to have test cases for some successful situations, like installing `https://libraries.io/github/aws/chalice` -- which depends on 35+ packages directly, including botocore, which has 900+ releases.

pradyunsg:
I suggest creating a script that fetches metadata from PyPI's JSON API for a package for all versions of that package, and dumps those details into a file (skipping packages without dependency metadata from the API). Then running this script for with all packages that get installed when you do `pip install chalice` will give you all the information needed to construct the test case.

uranusjr:
> I suggest creating a script that fetches metadata from PyPI's JSON API for a package for all versions of that package, and dumps those details into a file

Many of Monilino’s test cases are built this way (from RubyGems), and I’m 100% on board. It wouldn’t be a terrible idea to read from wheels even IMO since the spec generation is a one-off event.

pradyunsg:
https://github.com/pypa/pip/issues/7096

^ another simple-but-interesting case. :)

uranusjr:
Note: There are some cases raised in pypa/pip#7406 as well.

pradyunsg:
You can probably directly use https://github.com/chrahunt/packages. :)

uranusjr:
One more pypa/pip#7714

uranusjr:
(BTW I started implementing this in the `pypi-tests` branch.)

pradyunsg:
File a draft PR for the WIP branch?

pradyunsg:
One place worth "mining" is... pypa/pip#988. A *lot* of issues related to dependency resolution on pip's issue tracker were closed by cross-linking to that issue -- if you're looking for additional test cases, that'd be a good place to dig up for them.

uranusjr:
List of cases to pull (growing):

* [x] chalice
* [ ] Two extras required separately for one package (pypa/pip#7096)
* [ ] Two entries for one package, on with extra and one without (pypa/pip#7714)
* [ ] pypa/pip#7406
    * TODO: Read the thread
* [ ] pypa/pip#988
    * TODO: Read the thread

uranusjr:
To reproduce the virtualenv + pre-commit error:

```
[...]

PR #28, opened 2020-02-11
Prepare for Python test fixtures

uranusjr:
Toward #27.

This adds a script to scrap wheel metadata on PyPI to “freeze” dependency information for test cases.


uranusjr:
Alright, the script seems to work (albeit slow and with several hacks to make typing work), I’ll start adding test cases.

uranusjr:
Python packages have surprisingly large number of dependencies (if we include intrinsic) if we count all package extras. A large portion of this problem is probably contributed by Python lacking a dedicated way to specify a project’s test/dev dependencies.

pradyunsg:
Yup yup. At some point, we started suggesting that users use extra for development and testing dependencies. FWIW the resolver will have to resolve these instances too, so they are valid tests.

In some ways, the resolver should be doing what literally this script is doing, but "smarter".

uranusjr:
Indeed. The script is left “dumb” intentionally to ensure the resolver implementation does the right thing. My comment was more about complaining the incredible size of the frozen index (compared to RubyGems).

uranusjr:
Cool, seems like this thing is working, one test case down. Merging this and continue working on other fixtures.

PR #30, opened 2020-02-15
More PyPI tests

uranusjr:
Continuing #27.


tests/functional/python/test_resolvers_python.py
```diff
@@ -102,8 +112,22 @@ def get_dependencies(self, candidate):
 CASE_NAMES = [name for name in os.listdir(CASE_DIR) if name.endswith(".json")]
 
 
+XFAIL_CASES = {
+    "different-extras.json": "Resolver stalled",
+    "pyrex-1.9.8.json": "Resolver stalled",
```
pradyunsg: Interesting. What do you mean by stalled here?

uranusjr: It’s repeated trying one same combination. It shouldn’t, of course, so something’s wrong marking paths the resolver should not revisit. m

PR #34, opened 2020-03-27
Correctly handle name canonicalization in example

### LINKED ISSUE #33: resolution fails for 'django'
On current master (rev. 243acc576eae9f8bbd2fafa25653f3a7db2efc52).
Executing `pypi_wheel_provider.py django` fails with:
`resolvelib.resolvers.ResolutionTooDeep: 100`

uranusjr:
Previous, the example provider does not handle [name normalization](https://www.python.org/dev/peps/pep-0301/) correctly. Packages using an unnormalized name as canonical would send the resolver chasing its own tail.

This commit fixes the example to correctly normalize package names. The resolver fix will come later.

Fix #33.

### DIFF
--- a/examples/pypi_wheel_provider.py
+++ b/examples/pypi_wheel_provider.py
@@ -10,6 +10,7 @@ import html5lib
 from packaging.specifiers import SpecifierSet
 from packaging.version import Version, InvalidVersion
 from packaging.requirements import Requirement
+from packaging.utils import canonicalize_name
 
 from extras_provider import ExtrasProvider
 
@@ -18,13 +19,18 @@ PYTHON_VERSION = Version(platform.python_version())
 
 class Candidate:
     def __init__(self, name, version, url=None, extras=None):
-        self.name = name
+        self.name = canonicalize_name(name)
         self.version = version
         self.url = url
         self.extras = extras
 
         self._metadata = None
         self._dependencies = None
+
+    def __repr__(self):
+        if not self.extras:
+            return f"{self.name}=={self.version}"
+        return f"{self.name}[{','.join(self.extras)}]=={self.version}"
 
     @property
     def metadata(self):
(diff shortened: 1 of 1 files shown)


uranusjr:
(Not merging yet because it’s easier for me to use the broken example to figure out what’s going on in the resolver.)

pradyunsg:
I wonder if there's a good way to avoid canonical-names vs non-canonical-name by changing attribute names... Or something similar. We should probably do something in pip as well, to mitigate facing similar issues there too?

Maybe having `display_name` and `canonical_name` and getting rid of `name` would be better?

uranusjr:
I would be most happy if `packaging.requirement.Requirement` canonicalise `name` by default, and hold the non-normalised name in another attribute (e.g. `original_name`). But that’s just me, maybe that would break something I am not aware of.

pfmoore:
I just deleted a big post about "strong typing can do this" 🙂 

Basically it would be nice if we could take a "canonicalise at the boundary and always use canonical names internally" approach. But the need to display the original name makes that hard. And I really don't want to have both attributes everywhere, especially as they are usually properties. Too much boilerplate obscuring the main code for me.

pradyunsg:
Yet another option is to have something like pip's `HiddenText` object, to hold the variations in the name: `self.name.canonical` and `self.name.original`.

pfmoore:
My deleted comment suggested having a mypy type alias for `str` that tracked canonicalised names as a separate type (a bit like Rust handles things like OsStr and Path). But yes, an actual type might be good too.

IMO, it's borderline on over-engineering things, but we do hit the issue often enough that I suspect it's a worthwhile idea. Question - are you just suggesting this for the resolvelib example provider, or for pip as well (it would be a much more widespread change for pip)?

uranusjr:
I don’t think type aliasing works 😩 AFAIK Mypy treats type aliases as equivalent and will let it pass. We need to have an equivalent of single-field struct in C instead. And that would definitely be over-engineering IMO.

uranusjr:
I came to merge this since I’m done with the debugging. Let’s continue the discussion here though.

PR #38, opened 2020-04-02
Add extra reporter hooks to track requirements, pins and backtracks

pfmoore:


src/resolvelib/resolvers.py
```diff
@@ -224,6 +226,7 @@ def _backtrack(self):
 
             # Retract the last candidate pin, and create a new (b).
             name, candidate = self._states.pop().mapping.popitem()
+            self._r.backtrack(candidate)
```
uranusjr, line 229: Would `backtracking` be more consistent?

src/resolvelib/reporters.py
```diff
@@ -22,3 +22,15 @@ def ending_round(self, index, state):
     def ending(self, state):
         """Called before the resolution ends successfully.
         """
+
+    def adding_requirement(self, requirement):
+        """Adding a new requirement into the resolve criteria.
+        """
```
pradyunsg, line 28: nit: Documentation string phasing consistency would be nice.

src/resolvelib/resolvers.py
pradyunsg: It would be correct; given that the method is called `backtracking`. ;P

pfmoore: Point taken :-)

src/resolvelib/reporters.py
pfmoore: Not a nit, it's an important point and one I've just not go round to. Will fix this.

src/resolvelib/resolvers.py
pfmoore: Sigh. Simple examples don't backtrack. Thanks for spotting this :-)

uranusjr: I’ve been thinking about it for a while, this could be the tipping point I start adding type hints to the code base.

PR #40, opened 2020-04-03
Add a noxfile to make running lint and tests locally a bit easier

pfmoore:


pfmoore:
Thanks. I considered asking whether you preferred nox or tox, but nox was easier for me to set up, and it seemed like a debate was more trouble than it'd be worth 🙂

uranusjr:
In fact I do prefer Nox, so that makes things easy 😉

PR #42, opened 2020-04-09
Return parent information in ResolutionImpossible exception

pfmoore:


src/resolvelib/resolvers.py
```diff
@@ -279,8 +280,10 @@ def resolve(self, requirements, max_rounds):
                 name, crit = self._merge_into_criterion(r, parent=None)
             except RequirementsConflicted as e:
                 # If initial requirements conflict, nothing would ever work.
-                conflict_reqs = list(e.criterion.iter_requirement()) + [r]
-                raise ResolutionImpossible(conflict_reqs)
+                conflicts = [RequirementInformation(r, None)]
+                for req in e.criterion.iter_requirement():
+                    conflicts.append(RequirementInformation(req, None))
+                raise ResolutionImpossible(conflicts)
```
uranusjr, line 286: I think we can do `e.criterion.information + [RequirementInformation(r, Non)]`?

tests/functional/cocoapods/test_resolvers_cocoapods.py
```diff
@@ -163,7 +163,7 @@ def provider(request):
 
 
 def _format_conflicts(exc):
-    return {r.name for r in exc.requirements}
+    return {r.name for (r, p) in exc.causes}
```
uranusjr, line 166:
```suggestion
    return {r.name for r, _ in exc.causes}
```

It’s more obvious the second argument is intentionally unused IMO.

uranusjr:
This will need a news fragment as well because the interface changed. It’s probably worthwhile to have a `requirements` property as well? That one I’m not entirely sure.

pfmoore:
Sure, I'll add a news fragment. Is `42.feature.rst` the correct filename to use? (This is probably a feature rather than a bugfix).

I'm also not sure about a `requirements` property. I'm fine to add it, but it's not that hard to get the requirements from the causes and other than the tests, I'm not aware how many users this library has other than pip, so I can't really say if there's a real use case for "just the requirements". How important is backward compatibility at the moment?

uranusjr:
I’m not sure either. I know there are some users, but the number is likely still low enough we can affordd a breaking change.

Let’s not add the property for now and see how long we can get away with this. We can always add it if there are too many complaints.

Issue #39, opened 2020-04-03
Lazy-evaluate `is_satisfied_by`


uranusjr:
Context: https://github.com/pypa/pip/issues/7966#issuecomment-608100644

I *think* it is *theoratically* possible for the resolver to not process the candidate list immediately, but to simply add a “filter” thing on it, and check `is_satisfied_by` on a candidate only if it ever needs it (because all candidates with higher preferences don’t work). This would probably make the things a lot of complex than they are right now, but it sounds like a good thing for performance as well.

uranusjr:
c/f pypa/pip#8023

uranusjr:
We went a different route instead, by making `find_matches()` return lazy sequences instead.

PR #49, opened 2020-04-28
Introduce iter_matches() to replace find_matches()

### LINKED ISSUE #48: Resolving depends on the order of the root requirements in some cases
The following test fails.

```python
def test_requirements_different_candidate_sets():
    requirements = {
        "r1": ["c1"],
        "r2": ["c2"],
    }

    class Provider(AbstractProvider):
        def identify(self, d):
            return "r" if d.startswith("r") else d

        def get_preference(self, *_):
            return 0

        def get_dependencies(self, _):
            return []

        def find_matches(self, r):
            return requirements.get(r, [])

        def is_satisfied_by(self, r, c):
            if r == "r1":
                return c in requirements[r]
            elif r == "r2":
                # r2 accepts anything, even stuff it doesn't
                # return in find_matches
                return True
            return False

    resolver = Resolver(Provider(), BaseReporter())
    result = resolver.resolve(["r1", "r2"])
    assert result.mapping["r"] == "c1"

    resolver = Resolver(Provider(), BaseReporter())
    result = resolver.resolve(["r2", "r1"])
    assert result.mapping["r"] == "c1"
```

Note that the two tests at the end are the same request, just with the order of the requirements reversed.

The requirements have been constructed so that:

1. They are both for the same "project".
2. r1 is a normal sort of requirement.
3. r2 accepts the candidate r1 returns as well as its own one, but *doesn't* return that candidate in `find_matches`.

This specific configuration came up in https://github.com/pypa/pip/pull/8136.

I don't think the result of a `resolve()` call should be order-dependent like this, even if we argue that the provider is pathological. If we don't want to support providers like this, we need to document precisely what rules the provider has to follow - and we should validate them and produce a proper error, not just return an incorrect result.

(Personally, I'd like to get this case to work, as it's far easier to implement pip's constraints if we can).

uranusjr:
WIP, fix #48.

I immediately found a problem with my previous `find_all_matches(identifier)` approach: direct URL requirements. The idea of “all” matches is wrong by itself, since there is no universal set of candidates for a givan identifier.

The approach presented here instead simply presents all requirements known to the resolver at a given point, and ask the provider to return candidates that match all of them. This avoids the universal set problem (every candidate discovery is bounded by at least one requirement).

A side effect is this makes the provider a bit more difficult to implement without hitting the finder over and over again. We probably need to implement some kind of memoisation so multiple searches for the package only hit an index once.

### DIFF
--- a/examples/pypi_wheel_provider.py
+++ b/examples/pypi_wheel_provider.py
@@ -123,19 +123,20 @@ class PyPIProvider(ExtrasProvider):
     def get_preference(self, resolution, candidates, information):
         return len(candidates)
 
-    def find_matches(self, requirement):
+    def iter_matches(self, requirements):
         candidates = []
-        name = requirement.name
-        extras = requirement.extras
+        name = requirements[0].name
+        extras = requirements[0].extras
 
         # Need to pass the extras to the search, so they
         # are added to the candidate at creation - we
         # treat candidates as immutable once created.
         for c in get_project_from_pypi(name, extras):
             version = c.version
-            if version in requirement.specifier:
+            if all(version in r.specifier for r in requirements):
                 candidates.append(c)
-        return candidates
+        for c in reversed(candidates):
+            yield c
 
     def is_satisfied_by(self, requirement, candidate):
         if canonicalize_name(requirement.name) != candidate.name:
--- a/examples/pypi_wheel_provider.py
+++ b/examples/pypi_wheel_provider.py
@@ -126,8 +126,9 @@ class PyPIProvider(ExtrasProvider):
 
     def iter_matches(self, requirements):
         assert requirements, "resolver promises at least one requirement"
-        assert not any(r.extras for r in requirements[1:]), \
-            "extras not supported in this example"
+        assert not any(
+            r.extras for r in requirements[1:]
+        ), "extras not supported in this example"
 
         name = canonicalize_name(requirements[0].name)
 
--- a/examples/pypi_wheel_provider.py
+++ b/examples/pypi_wheel_provider.py
@@ -1,9 +1,10 @@
-import platform
+from email.message import EmailMessage
+from email.parser import BytesParser
 from io import BytesIO
+from operator import attrgetter
+from platform import python_version
 from urllib.parse import urlparse
 from zipfile import ZipFile
-from email.parser import BytesParser
-from email.message import EmailMessage
 
 import requests
 import html5lib
@@ -14,7 +15,7 @@ from packaging.utils import canonicalize_name
 
 from extras_provider import ExtrasProvider
 
-PYTHON_VERSION = Version(platform.python_version())
+PYTHON_VERSION = Version(python_version())
 
 
 class Candidate:
@@ -124,18 +125,21 @@ class PyPIProvider(ExtrasProvider):
         return len(candidates)
 
     def iter_matches(self, requirements):
-        candidates = []
-        name = requirements[0].name
-        extras = requirements[0].extras
+        assert requirements, "resolver promises at least one requirement"
+        assert not any(r.extras for r in requirements[1:]), \
+            "extras not supported in this example"
+
+        name = canonicalize_name(requirements[0].name)
 
         # Need to pass the extras to the search, so they
         # are added to the candidate at creation - we
         # treat candidates as immutable once created.
-        for c in get_project_from_pypi(name, extras):
+        candidates = []
+        for c in get_project_from_pypi(name, set()):
             version = c.version
             if all(version in r.specifier for r in requirements):
                 candidates.append(c)
-        for c in reversed(candidates):
+        for c in sorted(candidates, key=attrgetter("version"), reverse=True):
             yield c
 
     def is_satisfied_by(self, requirement, candidate):
--- a/examples/pypi_wheel_provider.py
+++ b/examples/pypi_wheel_provider.py
@@ -124,7 +124,7 @@ class PyPIProvider(ExtrasProvider):
     def get_preference(self, resolution, candidates, information):
         return len(candidates)
 
-    def iter_matches(self, requirements):
+    def find_matches(self, requirements):
         assert requirements, "resolver promises at least one requirement"
         assert not any(
             r.extras for r in requirements[1:]
@@ -140,8 +140,7 @@ class PyPIProvider(ExtrasProvider):
             version = c.version
             if all(version in r.specifier for r in requirements):
                 candidates.append(c)
-        for c in sorted(candidates, key=attrgetter("version"), reverse=True):
-            yield c
+        return sorted(candidates, key=attrgetter("version"), reverse=True)
 
     def is_satisfied_by(self, requirement, candidate):
         if canonicalize_name(requirement.name) != candidate.name:
--- a/src/resolvelib/providers.py
+++ b/src/resolvelib/providers.py
@@ -49,16 +49,16 @@ class AbstractProvider(object):
         raise NotImplementedError
 
     def iter_matches(self, requirements):
-        """Find all possible candidates that satisfy all given requirements.
+        """Find all possible candidates that satisfy the given requirements.
 
         This should try to get candidates based on the requirements' types.
         For VCS, local, and archive requirements, the one-and-only match is
         returned, and for a "named" requirement, the index(es) should be
         consulted to find concrete candidates for this requirement.
 
-        :param requirements: A collection of requirements to filter out
-            candidates from. All requirements are guarenteed to have the
-            same identifier. The collection is never empty.
+        :param requirements: A collection of requirements which all of the the
+            returned candidates must match. All requirements are guarenteed to
+            have the same identifier. The collection is never empty.
         :returns: An iterator that generates candidates by preference order,
             e.g. the most preferred candidate should come first.
         """
--- a/src/resolvelib/providers.py
+++ b/src/resolvelib/providers.py
@@ -48,7 +48,7 @@ class AbstractProvider(object):
         """
         raise NotImplementedError
 
-    def iter_matches(self, requirements):
+    def find_matches(self, requirements):
         """Find all possible candidates that satisfy the given requirements.
 
         This should try to get candidates based on the requirements' types.
@@ -59,8 +59,8 @@ class AbstractProvider(object):
         :param requirements: A collection of requirements which all of the the
             returned candidates must match. All requirements are guarenteed to
             have the same identifier. The collection is never empty.
-        :returns: An iterator that generates candidates by preference order,
-            e.g. the most preferred candidate should come first.
+        :returns: An iterable that orders candidates by preference, e.g. the
+            most preferred candidate should come first.
         """
         raise NotImplementedError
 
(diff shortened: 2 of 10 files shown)


pradyunsg:
Two birds with one stone - I like it. :)

pfmoore:
I'm not entirely sure I see the benefit of making `iter_matches` take a list of requirements. It makes the provider interface harder to implement and potentially less efficient, as you say, but I don't really see what the gain is (for the provider, specifically).

The "find all candidates for an identity" version had the benefit for the provider that it separated out "get candidates" and "filter candidates" into two disjoint operations, and put "get candidates" on the identity, where it (logically) belongs. It didn't fail because it was the wrong approach, just because "get all candidates for an identity" is impossible to actually implement in the real-world situation that pip implements<sup>1</sup>.

I don't see a similar benefit for the new version. What am I missing?

(Making the method an iterator, and reversing the order, on the other hand, I'm fine with. For the provider that makes it logically "yield results starting with the best option and getting more desparate as you go on", which fits how I think of interfaces like that very well 🙂)

<sup>1</sup> It's a consequence of not having all the metadata available immediately, like so many of the issues pip has :-(

uranusjr:
The problem is that “all candidates for an identify” means different things for `foo >= version` and `foo @ url`. If we just pass in `foo` there is no way to tell.

pfmoore:
Yes, I get that. But how does changing from passing in one *requirement* (not an identity) to many help with that? To put it another way, what code will be simplified by this change (I don't see any code in resolvelib itseld that's much improved) or what bug will be fixed?

I feel like I've forgotten some important bit of context here 🙁

uranusjr:
Currently (in master) the resolver basically does this:

```python
candidates = find_matches(req1)

# Afterwards...
candidates = [c for c in candidates is is_satisfied_by(req2, c)]
```

The problem in #48 from what I understand is that `find_matches()` and `is_satisfied_by()` may be implemented in a non-exchangable way, and you get different results if you flip `req1` and `req2`.

This change of interface essentially punts the problem to the provider, changing the resolver to

```python
candidates = list(iter_matches([req1]))

# Afterwards...
candidates = list(iter_matches([req1, req2]))
```

So there are not two different functions called depending on the order, but always the same one with everything the resolver knows at a given point.

I believe (not entirely sure) this also solves a problem where a direct URL candidate is added later during the resolution. Previously if `req1` is `foo` but `req2` is `foo @ https://example.com/foo-1.0.tar.gz`, the resolver would fail at the `is_satisfied_by()` call because there’s no way to “add” a candidate to the found matches. With the new interface, the provider can simply adds the direct URL at the second `iter_matches` call.

pfmoore:
Ah, OK. I see - that's the purpose of the change at around line 101 of resolvers.py. That makes sense to me. I hadn't seen how this addressed #48 because I'd not appreciated what that specific change was doing.

I'm not in love with the new API (it makes life more difficult for providers that want to be efficient, as you say, and in particular it makes the "obvious" provider implementation use a package index inefficiently) but I don't see a better answer.

examples/pypi_wheel_provider.py
```diff
@@ -123,19 +123,20 @@ def get_base_requirement(self, candidate):
     def get_preference(self, resolution, candidates, information):
         return len(candidates)
 
-    def find_matches(self, requirement):
+    def iter_matches(self, requirements):
         candidates = []
-        name = requirement.name
-        extras = requirement.extras
+        name = requirements[0].name
```
pfmoore, line 128: Add a comment here noting that the API guarantees that name is the same for all requirements?

examples/pypi_wheel_provider.py
```diff
@@ -123,19 +123,20 @@ def get_base_requirement(self, candidate):
...
+        extras = requirements[0].extras
```
pfmoore, line 129: This will break if the various requirements have different extras. I'm not sure if that's possible (as the extras are included in the name) but again it's a complexity that should be called out in a comment, at least.

examples/pypi_wheel_provider.py
```diff
@@ -123,19 +123,20 @@ def get_base_requirement(self, candidate):
...
         # Need to pass the extras to the search, so they
         # are added to the candidate at creation - we
         # treat candidates as immutable once created.
         for c in get_project_from_pypi(name, extras):
             version = c.version
-            if version in requirement.specifier:
+            if all(version in r.specifier for r in requirements):
                 candidates.append(c)
-        return candidates
+        for c in reversed(candidates):
```
pfmoore, line 138: Maybe do the `reversed()` on the call to `get_project_from_pypi`? That call itself depends on the order of elements in the index, so maybe `sorted(get_project_from_pypi(name, extras), key=operator.attrgetter('version'), reversed=True)` would be better still. Then we can just yield c in the loop, rather than building a list of candidates.

src/resolvelib/providers.py
```diff
@@ -48,17 +48,19 @@ def get_preference(self, resolution, candidates, information):
         """
         raise NotImplementedError
 
-    def find_matches(self, requirement):
-        """Find all possible candidates that satisfy a requirement.
+    def iter_matches(self, requirements):
+        """Find all possible candidates that satisfy all given requirements.
```
pfmoore, line 52:
```suggestion
        """Find all possible candidates that satisfy the given requirements.
```
Reads better without the repeated "all" in the sentence.

src/resolvelib/providers.py
```diff
@@ -48,17 +48,19 @@ def get_preference(self, resolution, candidates, information):
...
+        This should try to get candidates based on the requirements' types.
         For VCS, local, and archive requirements, the one-and-only match is
         returned, and for a "named" requirement, the index(es) should be
         consulted to find concrete candidates for this requirement.
 
-        The returned candidates should be sorted by reversed preference, e.g.
-        the most preferred should be LAST. This is done so list-popping can be
-        as efficient as possible.
+        :param requirements: A collection of requirements to filter out
+            candidates from. All requirements are guarenteed to have the
```
pfmoore, line 60: "A collection of requirements which all of the the returned candidates must match."

src/resolvelib/resolvers.py
```diff
@@ -98,11 +98,7 @@ def merged_with(self, provider, requirement, parent):
         """
         infos = list(self.information)
         infos.append(RequirementInformation(requirement, parent))
-        candidates = [
-            c
-            for c in self.candidates
-            if provider.is_satisfied_by(requirement, c)
-        ]
+        candidates = list(provider.iter_matches([r for r, _ in infos]))
```
pfmoore, line 101: It bothers me a bit that we always call list on `iter_matches`. I like the idea that a provider can yield results from `iter_matches` if it makes sense to do so, but the way we describe the method, I suspect people will thing that if they have built a list, they need to do

```python
for candidate in return_list:
    yield candidate
```

when actually this just means we break up and then rebuild the list 🙁 

I don't know what the term is for a function that returns "a list or something like an iterable or an iterator that can be converted to a list"... Surely there must be one? And what would be the type of such a beast (if we were using type annotations)? I feel like the typing people would know how to answer this.

tests/test_resolvers.py
```diff
@@ -23,8 +23,8 @@ def get_preference(self, *_):
         def get_dependencies(self, _):
             return []
 
-        def find_matches(self, r):
-            assert r is requirement
+        def iter_matches(self, rs):
+            assert rs[0] is requirement
```
pfmoore, line 27: `assert len(rs) == 1`?

examples/pypi_wheel_provider.py
uranusjr: I have added a line in the `AbstractProvider.iter_matches()` docstring that says *the collection is never empty*.

examples/pypi_wheel_provider.py
uranusjr: The extras will always match for pip, but may not in this example. I’m not entirely sure whether it’s worth it to implement something as elaborated here though; maybe I’ll rewrite the examples into one after finishing this PR.

examples/pypi_wheel_provider.py
uranusjr: Going with the `sorted()` approach.

src/resolvelib/resolvers.py
uranusjr: I’m torn on this. On one hand it is indeed typical for a provider to form a list at some point anyway, but there are also instances (e.g. pip’s!) where it’s easier to implement an iterator. And I also have some potential optimisation ideas to avoid repeating `list()` calls at some point (#39), so I want to avoid the additional `list()` call in the provider as well.

Maybe I’ll allow returning *either* an iterator or sequence. Reading [`collections.abc`](https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes), it’s possible to check `isinstance(candidates, Sequence)` to determine what is what. And we’ll need to rename the function again because it’s no longer returning an iterator.

pfmoore: Cool. That's what I was trying to suggest. (And looking at the commit text, yes, "iterable" was the term I couldn't remember, doh).

examples/pypi_wheel_provider.py
pfmoore: Works for me. In general, though, I do think this might make extras more complex to handle in the general case, and having a worked example somewhere would be good.

src/resolvelib/providers.py
```diff
@@ -48,17 +48,19 @@ def get_preference(self, resolution, candidates, information):
...
+        This should try to get candidates based on the requirements' types.
         For VCS, local, and archive requirements, the one-and-only match is
         returned, and for a "named" requirement, the index(es) should be
         consulted to find concrete candidates for this requirement.
 
-        The returned candidates should be sorted by reversed preference, e.g.
-        the most preferred should be LAST. This is done so list-popping can be
-        as efficient as possible.
+        :param requirements: A collection of requirements which all of the the
+            returned candidates must match. All requirements are guarenteed to
```
pfmoore, line 60:
```suggestion
            returned candidates must match. All requirements are guaranteed to
```

frostming:
Curious about what this PR does. IMHO it will lead to multiple PyPI requests on the same package during the resolution? Correct me if I am wrong. And I don't see the benefit of returning iterable since it is converted to a list anyway.

The key point here is to make it possible to add more candidates to the previous candidates set returned by `find_matches()`, I prefer an interface like
```python
def contribute_to_candidates(self, prev_candidates: Iterable[Candidate], req: Requirement) -> Iterable[Candidate]
```

uranusjr:
> IMHO it will lead to multiple PyPI requests on the same package during the resolution?

If the provider implementation is naive, yes. But it can perform very simple caching (the return value won’t change for a given package) to avoid that.

I did consider passing the previous candidates, but it is not obvious how the provider should handle the `prev_candidates` argument. What should the resolver pass in for the first ever call (when it has no knowledge on candidates)? If it’s `None` or an empty iterable, the provider will need to implement different logic for that, and it’s easier to get it wrong. In the end I decided it’s better to let the provider handle this instead:

```python
def iter_matches(reqs):
    candidates = finder.find_all_candidates(reqs[0].name)  # <-- Cache this function.
    seen_versions = set()

    installed_candidate = find_satisfying_installation(reqs)
    if installed_candidate:
        seen_versions.add(installed_candidate.version)
        yield installed_candidate
    for c in candidates:
        if c.version in seen_versions:
            continue
        seen_versions.add(c.version)
        yield c
```

frostming:
Thanks for the explanation, it makes much sense now.

PR #50, opened 2020-04-29
Add `parent` argument to `Reporter.adding_requirement`

### LINKED ISSUE #46: Make it possible (for a reporter) to determine which candidate a requirement originates from
This would be useful for reporters, to determine why a certain requirement is added. 

I'm imagining a `evaluating` hook, that's called at the start of the loop on `_attempt_to_pin_criterion`, prior to `pinning` but unconditionally (so that it doesn't actually have to match the criterion).

As an example, I want to be able to visualize false-starts / not-compatible-version checks, before they're pinned and currently, there's no way for the reporter to get this information.

pradyunsg:
Closes #46

### DIFF
--- a/examples/reporter_demo.py
+++ b/examples/reporter_demo.py
@@ -68,7 +68,7 @@ class Provider(resolvelib.AbstractProvider):
 
 
 class Reporter(resolvelib.BaseReporter):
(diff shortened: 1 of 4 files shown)


uranusjr:
Needs a news fragment :p

PR #53, opened 2020-05-23
Drop the overloading of the "dependency" term

pradyunsg:
This term has multiple conflicting definitions in this document.

In Candidate:

> Both of "requirement" and "dependency", however, SHOULD NOT be used when describing a Candidate, to avoid confusion.

In Requirement:

> A "dependency", if not clarified otherwise, also refers to this concept.

In this definition:

> A dependency can be either a requirement, or a candidate. In implementations you can treat it as the parent class and/or a protocol of the two.

---

I've opted to remove the single definition at the end, to give us nice and clean "dependency = requirement" naming. We'd likely need to change the variable name in `provider.identify(...)` to something better.

I propose just using `requirement_or_candidate` as the variable name, and avoiding this "pick a name" problem entirely. :P

### DIFF
--- a/examples/extras_provider.py
+++ b/examples/extras_provider.py
@@ -25,8 +25,8 @@ class ExtrasProvider(AbstractProvider):
     """A provider that handles extras.
     """
 
-    def get_extras_for(self, dependency):
-        """Given a dependency, return its extras.
+    def get_extras_for(self, requirement_or_candidate):
+        """Given a requirement or candidate, return its extras.
 
         The extras should be a hashable value.
         """
@@ -39,9 +39,9 @@ class ExtrasProvider(AbstractProvider):
         """
         raise NotImplementedError
 
-    def identify(self, dependency):
-        base = super(ExtrasProvider, self).identify(dependency)
-        extras = self.get_extras_for(dependency)
+    def identify(self, requirement_or_candidate):
+        base = super(ExtrasProvider, self).identify(requirement_or_candidate)
+        extras = self.get_extras_for(requirement_or_candidate)
         if extras:
             return (base, extras)
         else:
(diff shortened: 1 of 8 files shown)


uranusjr:
We’ll need to change the argument name in provider as the same time.

pradyunsg:
Done!

examples/extras_provider.py
```diff
@@ -39,9 +39,9 @@ def get_base_requirement(self, candidate):
         """
         raise NotImplementedError
 
-    def identify(self, dependency):
-        base = super(ExtrasProvider, self).identify(dependency)
-        extras = self.get_extras_for(dependency)
+    def identify(self, requirement_or_candidate):
```
pfmoore, line 42: Eww, is that *really* the best name we can use here? I wasn't that keen on "dependency" but even so...

I've always had a problem talking about this because I don't know what to call things. In pip terms, it's basically the "name" of the "project" that the requirement or candidate refers to, but neither of those terms are used by resolvelib, and I'm not sure what would be best as more neutral terms (or even if we need terms to be more neutral - at some point I feel that we're making it harder to reason about resolvelib because we're avoiding naming some of the concepts in order to be more generic...

But that's a lot of worrds for what is ultimately a fairly minor point, so feel free to ignore me...

pradyunsg: > Eww, is that _really_ the best name we can use here?

I mean, I'm sure we're open to other names. This is a actually-not-completely-horrible stop-gap solution until one of us has a wonderfully good brain wave. :P

uranusjr: I really should look into adding type hints… We could as well call this `item` (or whatever generic name) if it is type-hinted as `Union[Candidate, Requirement]`.

uranusjr:
Merging since the argument name isn’t really that important here (I don’t think the `identify` method would ever be needed to called with kwarg).

examples/extras_provider.py
pradyunsg: Heads up -- .pyi files (and py.typed) would likely be ideal. :)

PR #54, opened 2020-05-23
Reword the cautionary line about "specification" in Monillo

pradyunsg:
It seems to be an overtly specific reminder from back-when-we-were-looking-around-to-understand-what-to-do.

### DIFF
--- a/src/resolvelib/providers.py
+++ b/src/resolvelib/providers.py
@@ -13,7 +13,7 @@ class AbstractProvider(object):
         raise NotImplementedError
 
     def get_preference(self, resolution, candidates, information):
-        """Produce a sort key for given specification based on preference.
+        """Produce a sort key for given requirement based on preference.
 
         The preference is defined as "I think this requirement should be
         resolved first". The lower the return value is, the more preferred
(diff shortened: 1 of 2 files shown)


uranusjr:
IMO the line is still worthwhile, since [there are people using the term in Python packaging](https://discuss.python.org/t/structured-exchangeable-lock-file-format-requirements-txt-2-0/876/96?u=uranusjr). I wouldn’t object dropping the Monillo mention, however.

pradyunsg:
HEADS UP: we should update some documentation string in providers.py as well. Hold off on merging please!

pradyunsg:
I've skipped one instance in `provider.py` since I covered it in #53 (https://github.com/sarugaku/resolvelib/pull/53/files#diff-db65315aa1492e1d778ccbe3921352faR10)

src/resolvelib/providers.py
```diff
@@ -13,7 +13,7 @@ def identify(self, dependency):
         raise NotImplementedError
 
     def get_preference(self, resolution, candidates, information):
-        """Produce a sort key for given specification based on preference.
+        """Produce a sort key for given candidate based on preference.
```
uranusjr, line 16: This should be *requirement*, not candidate. (same below)

pradyunsg: Ahahahah

PR #55, opened 2020-05-30
Improve accessibiity for new-comers

McSinyx:
Hi, this fixes a few typos in documentations and add some docstrings to the PyPI wheel example.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -2,7 +2,7 @@
 ResolveLib
 ==========
 
-ResolveLib at the highest level provides a `Resolver` class that includes
+ResolveLib at the highest level provides a ``Resolver`` class that includes
 dependency resolution logic. You give it some things, and a little information
 on how it should interact with them, and it will spit out a resolution result.
 
@@ -41,11 +41,10 @@ The intention of this section is to unify the terms we use when talking about
 this code base, and packaging in general, to avoid confusion. Class and
 variable names in the code base should try to stick to terms defined here.
 
-Things passed into `Resolver.resolve()` and provided by the provider are all
+Things passed into ``Resolver.resolve()`` and provided by the provider are all
 considered opaque. They don't need to adhere to this set of terminologies.
 Nothing can go wrong as long as the provider implementers can keep their heads
 straight.
-
 
 Package
 -------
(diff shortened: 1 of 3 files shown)


README.rst
```diff
@@ -2,33 +2,33 @@
...
+ResolveLib at the highest level provides a ``Resolver`` class that includes
 dependency resolution logic. You give it some things, and a little information
 on how it should interact with them, and it will spit out a resolution result.
 
 
 Intended Usage
 ==============
 
-::
+.. code-block:: python
```
uranusjr, line 13: The readme intentionally only uses built-in RestructureText syntaxes since it may have compatibility issues when the project is uploaded. `.. code-block::` is a Sphinx extension.

README.rst
```diff
@@ -2,33 +2,33 @@
...
 
-    import resolvelib
+   from resolvelib import BaseReporter, Resolver
```
uranusjr, line 15: Why? I like the original qualified imports better than this.

examples/pypi_wheel_provider.py
```diff
@@ -3,6 +3,7 @@
 from io import BytesIO
 from operator import attrgetter
 from platform import python_version
+from sys import argv
```
uranusjr, line 6: It is a common pattern to only import sys in main if possible.

README.rst
McSinyx: GitHub and PyPI take `code-block` just fine, however I'm not entirely sure about other rendering engines.  A plus point for `code-block` is that some editors may allow syntax highlighting as well, e.g.
2020-05-31T11:05:57

That being said, I'll still revert the change since the highlighting we have here is minimal.

README.rst
McSinyx: I'll revert then since I don't really have a strong opinion on this one.

examples/pypi_wheel_provider.py
McSinyx: Thanks, TIL.

McSinyx:
Thanks for the review, I've just addressed the requested changes.

McSinyx:
Thanks for the quick merge!

PR #56, opened 2020-09-17
Pass ALL requirements to find_matches()

uranusjr:
For pypa/pip#8785. See https://github.com/pypa/pip/issues/8785#issuecomment-678885871 for context.

In order to know about `a @ ./a` when we run `find_matches` for `a[z]`, we need to somehow be able to access requirements from other packages. So this changes the arguments of `find_matches` again, and instead of passing a list of requirements, it now passes a mapping that contains all requirements. Thus we can do something like

```python
def find_matches(self, ident, all_reqs):
    reqs = all_reqs[ident]  # type: List[Requirement]
    for r in all_reqs.get(self.identify(reqs[0]), []):
        if r.is_explicit():
            return ExplicitCandidate(r)
    # Use index as we did...
```

This is posted so I can create a draft PR on pip to make sure things work.

### DIFF
--- a/examples/extras_provider.py
+++ b/examples/extras_provider.py
@@ -22,8 +22,7 @@ from resolvelib.providers import AbstractProvider
 
 
 class ExtrasProvider(AbstractProvider):
-    """A provider that handles extras.
-    """
+    """A provider that handles extras."""
 
     def get_extras_for(self, requirement_or_candidate):
         """Given a requirement or candidate, return its extras.
--- a/examples/pypi_wheel_provider.py
+++ b/examples/pypi_wheel_provider.py
@@ -124,7 +124,8 @@ class PyPIProvider(ExtrasProvider):
     def get_preference(self, resolution, candidates, information):
         return len(candidates)
 
-    def find_matches(self, requirements):
+    def find_matches(self, identifier, all_requirements):
+        requirements = all_requirements[identifier]
         assert requirements, "resolver promises at least one requirement"
         assert not any(
             r.extras for r in requirements[1:]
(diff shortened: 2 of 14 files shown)


pradyunsg:
Hmm... Won't this make implementing pubgrub more difficult?

pfmoore:
I think that yes, this is going to tie us (both here and in pip and other clients) more closely to a specific algorithm. That is pretty much unavoidable, we either tailor the API and implementation to the use cases, or we reject the use case from pip because it can't be implemented in an algorithm-independent way.

I don't know enough about PubGrub (yet!) to say whether this change is going to make switching to PubGrub impossible, but we've already lost the opportunity of using SAT because the API allows requirements to be computed "as needed". Isn't this just another case of the same sort of thing?

Personally, I think we *should* be looking at desupporting or changing the behaviour of some of the more awkward edge cases (hello, extras!) in favour of keeping resolvelib more flexible and general, but it's hard to see how we'd achieve that practically in pip, given the history of prioritising "do what people need" over "have a well-defined and logical model" 🙁 And like it or not, "support pip's use cases" is a key driver for resolvelib functionality.

uranusjr:
I didn’t really think through all the details (if I did I’d have written a PubGrub implementation by now 😛), but I *think* this is possible in PubGrub. I intentionally implemented the `all_requirements` argument as a custom object with a mapping-like interface, instead of an actual dict, so any implementation should be able to construct the requirements lazily when relevent keys are accessed.

Thinking this from the other side, pip is extremely unlikely to drop support to either URL specs and extras without significant backlash, and the resolution has to do some kind of “inspecting other package’s current requirement state” to make it possible for these to work together. If this makes it impossible to implement PubGrub, that’d likely mean that PubGrub is no-go for pip.

uranusjr:
This was reinvented as a part of #77 instead.

PR #57, opened 2020-10-02
Implement lazy candidate sequence class

uranusjr:
Pull in logic from pypa/pip#8932.

### DIFF
--- a/src/resolvelib/resolvers.py
+++ b/src/resolvelib/resolvers.py
@@ -1,8 +1,7 @@
 import collections
 
-from .compat import collections_abc
 from .providers import AbstractResolver
-from .structs import DirectedGraph
+from .structs import DirectedGraph, build_iter_view
 
 
 RequirementInformation = collections.namedtuple(
@@ -77,15 +76,10 @@ class Criterion(object):
     @classmethod
     def from_requirement(cls, provider, requirement, parent):
         """Build an instance from a requirement."""
-        candidates = provider.find_matches([requirement])
-        if not isinstance(candidates, collections_abc.Sequence):
-            candidates = list(candidates)
-        criterion = cls(
-            candidates=candidates,
-            information=[RequirementInformation(requirement, parent)],
-            incompatibilities=[],
-        )
-        if not candidates:
+        cands = build_iter_view(provider.find_matches([requirement]))
+        infos = [RequirementInformation(requirement, parent)]
+        criterion = cls(cands, infos, incompatibilities=[])
+        if not cands:
             raise RequirementsConflicted(criterion)
         return criterion
 
@@ -99,11 +93,9 @@ class Criterion(object):
         """Build a new instance from this and a new requirement."""
         infos = list(self.information)
         infos.append(RequirementInformation(requirement, parent))
-        candidates = provider.find_matches([r for r, _ in infos])
-        if not isinstance(candidates, collections_abc.Sequence):
-            candidates = list(candidates)
-        criterion = type(self)(candidates, infos, list(self.incompatibilities))
-        if not candidates:
+        cands = build_iter_view(provider.find_matches([r for r, _ in infos]))
+        criterion = type(self)(cands, infos, list(self.incompatibilities))
+        if not cands:
             raise RequirementsConflicted(criterion)
         return criterion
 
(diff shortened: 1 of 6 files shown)


McSinyx:
I'm ultimately amused by the fact that we (people who work with software in general) went from lazy evaluated Lisp cons cells to Python destructive iterators and back to `_Peekable`.  Jokes aside it's a pleasure for me to watch this very interface involving so elegantly.

uranusjr:
Someday I would propose adding `itertools.peekable()` since that’s just so generally useful. although I just realised that we don’t actually need it in this case.

pradyunsg (approved):
LGTM!

src/resolvelib/structs.py
```diff
@@ -62,3 +67,32 @@ def iter_children(self, key):
...
+    def __iter__(self):
+        return self._factory()
+
+    def excluding(self, candidate):
+        """Create a new ``Candidates`` instance excluding ``candidate``."""
+
+        def factory():
+            return (c for c in self._factory() if c != candidate)
+
+        return type(self)(factory)
```
McSinyx, line 98: Nitpick: I think just `return Candidates(c for c in self._factory() if c != candidate)` would fit the bill and is more readable.

src/resolvelib/resolvers.py
```diff
@@ -99,11 +95,11 @@ def merged_with(self, provider, requirement, parent):
...
-        candidates = provider.find_matches([r for r, _ in infos])
-        if not isinstance(candidates, collections_abc.Sequence):
-            candidates = list(candidates)
-        criterion = type(self)(candidates, infos, list(self.incompatibilities))
-        if not candidates:
+        cands = Candidates(provider.find_matches([r for r, _ in infos]))
+        criterion = type(self)(cands, infos, list(self.incompatibilities))
+        try:
+            next(iter(cands))
+        except StopIteration:
```
McSinyx, line 102: I suggest defining a `isempty` function to make this (and the one below) looks less complex.

McSinyx (commented):
This get more elegant every iteration!

src/resolvelib/structs.py
uranusjr: The extra function is needed, otherwise the iterator gets resolved eagerly and breaks the optimisation.

McSinyx: Fix me if I'm wrong, but it seems that if we feed `__init__` an iterator like so, then `self._factory()` would be `iter(matches)`.

uranusjr: Correct, but that’s no reusable because `iter(iterator) is iterator`. To make what you want work we need to invent an iterable class that caches the iterator’s content on iteration and return those cached values instead of the original iterator later on. That’s totally doable, but I didn’t like what I came up with and decided to not so that.

McSinyx: Ah, I see.

pfmoore (changes requested):
Looks good to me, just a couple of minor improvements I think are worth considering.

src/resolvelib/structs.py
```diff
@@ -62,3 +67,32 @@ def iter_children(self, key):
...
+    """
+
+    def __init__(self, matches):
+        if callable(matches):
+            self._factory = matches
+        elif not isinstance(matches, collections_abc.Sequence):
+            self._factory = functools.partial(iter, list(matches))
+        else:
+            self._factory = functools.partial(iter, matches)
+
```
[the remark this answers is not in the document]
uranusjr: `candidates` passed to `get_preference` isn’t a `Candidates` instance, but the underlying iterator, since I want to keep the class an implementation detail if possible. I also want to avoid a custom `__len__` since it be a performance trap for implementers when they provide an iterator.

OTOH, lazy evaluation isn’t always needed (in fact I’d argue Python package’s need of it is a niche case). All of the tests certainly don’t need it. So maybe it’d be better if we are able to pass the return value of `find_matches` transparently to `get_preference`, i.e. you get a list in `get_preference` if you provided the resolver a list, and an iterator if you provided an iterator. I’ll look into the possibility.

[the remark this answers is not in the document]
pfmoore: > candidates passed to get_preference isn’t a Candidates instance

Ah, sorry. I missed that...

uranusjr:
Continuing https://github.com/sarugaku/resolvelib/pull/57/files#r499134413, I re-implemented the `Candidates` wrapper into two distinct classes for the sequence and callable cases.

pfmoore (approved):

uranusjr:
For some reason GitHub does not let me resolve the `next(iter(cands))` comment, nor tag @McSinyx for review 😞

pradyunsg (approved):

PR #60, opened 2020-11-04
Carry all incompatibilities during backtracking

uranusjr:
Consider this situation:

* Candidate `Ax` depends on `B`
* Candidate `B1` depends on `C==1` `D==1` `E==1`
* Candidate `B2` depends on `C==2` `D==1` `E==1`
* Candidate `C1` depends on `D==1`
* Candidate `C2` depends on `D==1`
* Candidate `D1` depends on `E!=1`

In the previous implementation, the conflict on `E` is discovered after we resolved `Ax-B1-C1-D1`. `D1` is marked as an incompatibility, we backtrack the `C1` pin. But now we don’t have an available `C`, and need to also backtrack the `B1` pin. At this point, however, the previous implementation would fail to “remember” that `D1` is also marked as incompatible, and proceed to try `B2`. That would eventually fail, we backtrack to the same point, and end up trying `B1` again. Now we are stuck.

This fix uses a list to remember all the candidates marked as incompatible along the whole backtrack process, and “re-mark” them in parent states. This makes the resolver aware, when it backtracks `B1`, that `B2` is also not viable, and avoid hitting it.

### DIFF
--- a/src/resolvelib/resolvers.py
+++ b/src/resolvelib/resolvers.py
@@ -99,16 +99,16 @@ class Criterion(object):
             raise RequirementsConflicted(criterion)
         return criterion
 
-    def excluded_of(self, candidate):
-        """Build a new instance from this, but excluding specified candidate.
+    def excluded_of(self, incompatibilities):
+        """Build a new instance from this, but excluding specified candidates.
 
         Returns the new instance, or None if we still have no valid candidates.
         """
-        cands = self.candidates.excluding(candidate)
+        cands = self.candidates.excluding(incompatibilities)
         if not cands:
             return None
         incompats = list(self.incompatibilities)
-        incompats.append(candidate)
+        incompats.extend(incompatibilities)
         return type(self)(cands, list(self.information), incompats)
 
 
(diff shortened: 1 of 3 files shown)


webknjaz:
The solution sounds legit. Do you have a regression test case for this?

uranusjr:
I’m still trying to come up with a good, real world, test case. Both cases available in pypa/pip#9011 are quite big and I want to trim them down to a managable input that contains only PyPI wheels.

pradyunsg:
TBH, a synthetic test case (as described in OP) should also be sufficient.

uranusjr:
I tried to create a test case from the abstract example I described above, and… it passed on master. I think my diagnosis is on the right trck, and the fix *could* actually be correct (it does fix the real-world case?), but it’s very difficult to find the right input that hits the exact code path to trigger the issue. Maybe we need more candidates on a package? More packages hitting the conflict? A conflict deeper in the graph? All or some combination of the above? I don’t know. 😟

brainwane:
Per [a meeting today](https://wiki.python.org/psf/PackagingWG/2020-11-18-pip-team-meeting), as developers created a substantive and working test case, they found significant problems with their original approach, and so this issue needs more investigation and discussion to figure this out. This is a blocker (as is https://github.com/pypa/pip/issues/9011) to the pip 20.3 release.

I believe @uranusjr is poking at this a bit in his volunteer time -- is that correct? Tzu-ping?

@pradyunsg plans to finish the other remaining things on his 20.3 https://github.com/pypa/pip/issues/8936 checklist first, so he can make use of TP's investigations, finish and merge a fix, then immediately cut the 20.3 release.

uranusjr:
Yes, I’ve successfully reproduced the issue in #61 (which this PR would not fix) and plan to continue investigation and work on a fix there.

uranusjr:
Closing in favour of #61.

PR #61, opened 2020-11-18
Fix stuck resolution

uranusjr:
Eventually fix pypa/pip#9011. Currently only a test case that does not pass.


uranusjr:
Closing in favour of #62, which is basically the same thing, but with cleaner implementation (IMO) and better commits.

PR #62, opened 2020-11-27
Rework backtracking and state management

uranusjr:
For pypa/pip#9011.

I broke down the change into several commits. The last commit is the real "meat", others are only clean ups and test cases.

Quoting from the commit message:

The "forward" part of resolution now works like this:

* A "root state" is populated. This state does not produce a "pin" (decide on a candidate to satisfy a requirement), but only contain the root requirements, so subsequence rounds can use it as basis.
* Each subsequent round pushes a state, pin a candidate on it for it, and check whether the pin results in a conflict.
* If no conflcits occur, this round ends, and a new round is started. This happens until either all sets are pinned.

When a conflict happen, the backtracking process discards the current round's state, and try to modify the previous state so the conflict can be avoided. If this modification is unsuccessful, the still previous
state is tried, until a workable state is found (or we run out of choices).

This makes each round and its associating state's lifeline easier to reason with. Each successful forward round would create a state with a pin in the stack. A round ending with backtracking will not result in a new state in the stack, but have a previously-pinned state unpinned (and modified), with everything newer than it removed.


pradyunsg (approved):
Appreciate the comments added in! I'd been struggling to follow the code flow earlier, and these comments are immensely helpful!

pfmoore (approved):
I agree the comments make the logic a lot easier to follow. The new process looks good to me, I like the fact that each round the invariant is that the top stack item is in an unpinned state, that makes the reasoning much easier to follow.

The new structure gives me much more confidence that the logic is correct 🙂

PR #65, opened 2020-12-09
Backtrack if incompatibilities fail to apply

uranusjr:
For pypa/pip#9180. The closure function looks so ugly but I can’t come up with something better :(

### DIFF
--- a/src/resolvelib/resolvers.py
+++ b/src/resolvelib/resolvers.py
@@ -271,28 +271,36 @@ class Resolution(object):
                 for k, v in broken_state.criteria.items()
             ]
 
+            # Also mark the newly known incompatibility.
+            incompatibilities_from_broken.append((name, [candidate]))
+
             self._r.backtracking(candidate)
 
             # Create a new state from the last known-to-work one, and apply
             # the previously gathered incompatibility information.
-            self._push_new_state()
-            for k, incompatibilities in incompatibilities_from_broken:
-                try:
-                    crit = self.state.criteria[k]
-                except KeyError:
-                    continue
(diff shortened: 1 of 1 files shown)


pfmoore (approved):

pradyunsg:
I feel dumb -- I don't understand the need for the closure. :(

pfmoore:
I assumed it was just to name the block of code, for readability. If it's more than that, I'm as dumb as you, and I clicked "Approve" as well 🙂

uranusjr:
The closure is needed because we want to break a loop from a loop nested inside it. The high-level POV is

```python
while len(self._states) >= 3:
    ...
    for k, incompatibilities in incompatibilities_from_broken:
        ...
        if criterion is None:
            # We want to break the "while" loop here.
```

And a closure with early return is the best way I can think of to do it.

McSinyx:
I suppose we can use the following, but I'm not sure if it is any easier to read than a closure.

```python
while ...:
    for ...:
        if ...:
            break
    else:
        continue
    break
```

uranusjr:
@McSinyx I thought of using `else:` as well, but preferred the closure approach. Also we need to `continue` the outer `while`, and I couldn’t find a nice way to do it with `else:`.

pfmoore:
I noticed that after I posted my comment. IMO using the closure for the early exit is fine, the code is perfectly readable as it stands.

pradyunsg:
> I'm not sure if it is any easier to read than a closure.

FWIW, I would find this form easier to read, but only marginally, so :shrug:. This does look good to me now that I'm looking at it with that context! ^>^

PR #66, opened 2020-12-11
Allow resolution to skip backtracking

TheKevJames:
This change is meant to support pypa/pip#9258, eg. to allow users to opt
out of backtracking behaviour in cases where they would prefer to
receive failure or conflict results immediately, for example in cases
where they wish to use this resolver to help them improve their version
pinning strategies.

The full explanation is on the linked PR, though I'm of course happy to
accept more specific `resolvelib`-related feedback here as well!

### DIFF
--- a/src/resolvelib/resolvers.py
+++ b/src/resolvelib/resolvers.py
@@ -297,7 +297,7 @@ class Resolution(object):
         # No way to backtrack anymore.
         return False
 
-    def resolve(self, requirements, max_rounds):
+    def resolve(self, requirements, max_rounds, should_backtrack):
         if self._states:
             raise RuntimeError("already resolved")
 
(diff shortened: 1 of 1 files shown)


uranusjr:
I don’t think this requires a resolvelib change. To make the resolver to not backtrack, you can simply return only one candidate from `find_matches()`.

TheKevJames:
@uranusjr grabbing the relevant-to-this-repo part of my response https://github.com/pypa/pip/pull/9258#issuecomment-743474226 : 

I think the difference between returning fewer options in `find_matches()` vs what I've done here by disabling backtracking comes down to debuggability, in my mind. By reducing the number of options in `find_matches()`, we are preventing the resolver from gathering the potential `failure_causes` from that first iteration of resolving which can be incredibly useful in debugging; by simply disabling `backtracking` we are allowing the tool to do the heavy lifting of finding the potentially compatible versions for us and offloading only the decision of _how to resolve_ to the developer. If we instead return only a single option, the developer would not only need to determine how to resolve the issue but would additionally need to determine for themselves what the potential issue is in the first place.

Perhaps the flag is poorly named here, as per your comments in the other PR. Would renaming this more along the lines of `attempt_automated_conflict_resolution`, or the inverse of `fail_fast_on_conflict`, make more sense?

PR #67, opened 2021-01-05
Tree pruning

uranusjr:
This produces a ResolutionImpossible against `pip install prefect[all_extras] --only-binary :all:` in ~30s.

Still much to work on, mainly error reporting seems wrong. I can’t make sense of the report from the `perfect` example. It contains many `prefect[all-extras]` entries (different versions), but does not actually show where the conflict is. But the approach itself seems to work.


uranusjr:
Also need to write comments to describe how things work.

pradyunsg (commented):
Approach looks great to me.

I'm a little concerned about the flexibility we're giving the provider here, but, I guess that's a ship that sailed a while back. 🙃

src/resolvelib/resolvers.py
```diff
@@ -296,13 +325,13 @@ def _patch_criteria():
...
-                return True
+                return
 
             # State does not work after applying known incompatibilities.
             # Try the still previous state.
 
-        # No way to backtrack anymore.
-        return False
+        # No way to backtrack anymore. Give up.
+        raise ResolutionImpossible([i for c in causes for i in c.information])
```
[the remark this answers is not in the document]
uranusjr: Included accidentally. Removed.

src/resolvelib/providers.py
```diff
@@ -94,6 +94,21 @@ def get_dependencies(self, candidate):
...
+        """Whether the two given requirement sets find the same candidates.
+
+        This is used by the resolver to perform tree-pruning. If the two
+        requirement sets provide the same candidates, the resolver can avoid
+        visiting the subtree again when it's encountered, and directly mark it
+        as a dead end instead.
+
+        Both arguments are iterators yielding requirement objects. A boolean
+        should be returned to indicate whether the two sets should be treated
+        as matching.
```
uranusjr, line 107: This should also mention `requirements_a` is the new state, and `requirements_b` is from a known-to-fail state. This is important since the provider can implement range merging to return True when `requirements_a` is a *subset* of `requirements_b` to prune additional subtrees.

(The method name should probably change.)

uranusjr: OTOH if `requirements_a` is a superset of `requirements_b`, this should be used to exclude `requirements_b` from `requirements_a` so the resolver can avoid visiting some subtrees. Maybe the interface should be defined to perform specifier merging instead.

PR #70, opened 2021-01-18
New find_matches() arg to exclude candidates

### LINKED ISSUE #68: Candidate filtering depends on candidate equality logic implicitly
It should not. This currently works in pip because pip’s implementation caches candidates so equality works via object identity, and in tests because all tests currently use tuples (or namedtuples) to represent candidates.

Some potential solutions:

* Explicitly require the implementer to supply `Candidate.__eq__` (and `__ne__` on Python 2). I don’t like this.
* Add `Provider.is_equal_to(candidate_a, candidate_b)`.
* Add `Provider.excluding(candidate_iterable, excluding_candidates)`.

uranusjr:
Fix #68.


uranusjr:
Merging this change with #74.

PR #72, opened 2021-02-25
Add Typehint in the stub files

frostming:


uranusjr:
Thanks! I’ve been thinking about doing this for a while, but never quite got enough motivation to write all the generic boilerplate code.

uranusjr:
We should probably also add the check to `noxfile.py`.

Issue #74, opened 2021-03-07
Upcoming breaking changes


uranusjr:
I’m going to work toward 0.6.0, which is going to contain several breaking changes. This post serves both as a heads-up and summary to myself so I don’t forget something or have to add new things at the last minute.

### Keyword arguments

All Provider and Reporter methods will be called with keyword arguments from now on. This means you can start adding `**kwargs` for future compatibility now.

### `find_matches` argument changes

```python
def find_matches(
    identifier: Identifier,
    requirements: Mapping[Identifier, List[Requirement]],
    incompatibilities: Mapping[Identifier, List[Candidate]],
) -> Matches: ...
```

The current `requirements: List[Requirement]` can be replaced by `requirements[identifier]`. The new `incompatibilities` argument provides incompatible candidates already known to the resolver. The provider *must* exclude them from the result.

### `get_preference` argument changes

```python
def get_preference(
    identifier: Identifier,
    resolutions: Mapping[Identifier, Candidate],
    candidates: Mapping[Identifier, Collection[Candidate]],
    information: Mapping[Identifier, Sequence[Information]],
) -> Preference: ...
```

`Identifier` is the return type of `identify`. `Criterion` is the same as in `Result`.

Old `get_preference()` arguments can be replaced by:

* `resolution`: `resolutions.get(identifier)`.
* `candidates`: `candidates[identifier].candidates` is an abstraction around the return value of `find_matches()`. It implements the equence protocol if you returned a list from `find_matches()`. Otherwise, it is an iterable container.
* `information`: `criteria[identifier].information`.

PR #75, opened 2021-03-08
New get_preference() arguments

uranusjr:
Part of #74. Information about *other* dependencies are now passed to the provider. This allows better ordering logic that based on how "far" a dependency is from the root, instead of only able to distinguish transitivity.


uranusjr:
Splitting out the typing fixes first, will redo this from master again.

PR #81, opened 2021-07-29
start_backtracking suggestion

nadavwe:
resolvelib changes to solve https://github.com/pypa/pip/issues/10210.


tests/test_resolvers.py
```diff
@@ -91,3 +92,54 @@ def is_satisfied_by(self, requirement, candidate):
...
+        resolver = Resolver(Provider(), reporter)
+        try:
+            resolver.resolve(*args)
+            return reporter.backtracking_causes
+        except ResolutionImpossible as e:
+            return e.causes
+
+    backtracking_causes = run_resolver([("a", {1, 2}), ("b", {1})])
+    exception_causes = run_resolver([("a", {2}), ("b", {1})])
+    assert exception_causes == backtracking_causes
```
nadavwe, line 145: I wanted to make it obvious that the backtracking causes are exactly the same as the failure when the "best" version is selected.
@uranusjr do you prefer i give an explicit value instead?

uranusjr:
I’m +1 to the idea, but we should use a different function name. It’s not at all clear what the difference between `start_backtracking` and `backtracking` is (also the tense is inconsistent to other functions).

Maybe we can call this something like `resolving_conflicts`?

src/resolvelib/reporters.py
```diff
@@ -30,8 +30,9 @@ def adding_requirement(self, requirement, parent):
             requirements passed in from ``Resolver.resolve()``.
         """
 
-    def start_backtracking(self, causes):
-        """Called when backtracking started because of collision in requirements
+    def resolving_conflicts(self, causes):
+        """Called when a collision in requirements was detected and
+        an attempt to resolve them was started
```
uranusjr, line 35:
```suggestion
    def resolving_conflicts(self, causes):
        """Called when starting to attempt requirement conflict resolution.
```

Let's keep this succinct.

src/resolvelib/resolvers.py
```diff
@@ -369,7 +369,7 @@ def resolve(self, requirements, max_rounds):
                 causes = [i for c in failure_causes for i in c.information]
                 # Backtrack if pinning fails. The backtrack process puts us in
                 # an unpinned state, so we can work on it in the next round.
-                self._r.start_backtracking(causes)
+                self._r.resolving_conflicts(causes)
```
uranusjr, line 382:
```suggestion
                self._r.resolving_conflicts(causes=causes)
```

All provider and reporter calls should use named arguments for forward compatibility.

uranusjr (changes requested):
Some minor changes are needed.

tests/test_resolvers.py
uranusjr: This is better IMO if you want to express the causes should be the same.

uranusjr (approved):

uranusjr:
Could you rebase to or merge the latest master branch so we can apply the new GitHub Action checks?

uranusjr:
Linter error!

nadavwe:
> Linter error!

sorry for that... :/

uranusjr:
Nothing to be sorry about, the purpose of the CI being there to catch our errors 😛

nadavwe:
> Nothing to be sorry about, the purpose of the CI being there to catch our errors 😛

If I had run nox properly before committing, none of those problems would have occurred...
that's why I'm sorry, needless ping-pongs. 🏓 😳

anyway, I ran nox now and it passes locally, I hope that CI agrees with me. 🤞

uranusjr:
0.8.1 released with this change.

PR #84, opened 2021-09-16
Store failure cause and provide to _get_preference

notatallshaw:
As per https://github.com/pypa/pip/issues/10479

If pip prefers failure causes then resolvelib must provide them to get_preference

Required for https://github.com/pypa/pip/pull/10481

### DIFF
--- a/src/resolvelib/resolvers.py
+++ b/src/resolvelib/resolvers.py
@@ -234,6 +234,7 @@ class Resolution(object):
             # backtracking looks at this mapping to get the last pin.
             self.state.mapping.pop(name, None)
             self.state.mapping[name] = candidate
+
             return []
 
         # All candidates tried, nothing works. This criterion is a dead
--- a/src/resolvelib/resolvers.py
+++ b/src/resolvelib/resolvers.py
@@ -234,7 +234,6 @@ class Resolution(object):
             # backtracking looks at this mapping to get the last pin.
             self.state.mapping.pop(name, None)
             self.state.mapping[name] = candidate
-    
             return []
 
         # All candidates tried, nothing works. This criterion is a dead
--- a/src/resolvelib/resolvers.py
+++ b/src/resolvelib/resolvers.py
@@ -186,7 +186,7 @@ class Resolution(object):
                 self.state.criteria,
                 operator.attrgetter("information"),
             ),
-            backtrack_causes=self.state.backtrack_causes
+            backtrack_causes=self.state.backtrack_causes,
         )
 
     def _is_current_pin_satisfying(self, name, criterion):
@@ -338,10 +338,12 @@ class Resolution(object):
 
         # Initialize the root state.
         self._states = [
-            State(mapping=collections.OrderedDict(),
-                  criteria={},
-                  backtrack_causes=[])
-            ]
+            State(
+                mapping=collections.OrderedDict(),
+                criteria={},
+                backtrack_causes=[],
+            )
+        ]
         for r in requirements:
             try:
                 self._add_to_criteria(self.state.criteria, r, parent=None)
@@ -376,8 +378,8 @@ class Resolution(object):
                 # an unpinned state, so we can work on it in the next round.
                 success = self._backtrack()
                 self.state.backtrack_causes[:] = [
-                        i for c in failure_causes for i in c.information
-                    ]
+                    i for c in failure_causes for i in c.information
+                ]
 
                 # Dead ends everywhere. Give up.
                 if not success:
--- a/src/resolvelib/resolvers.py
+++ b/src/resolvelib/resolvers.py
@@ -375,7 +375,6 @@ class Resolution(object):
 
                 # Dead ends everywhere. Give up.
                 if not success:
-                    causes = [i for c in failure_causes for i in c.information]
                     raise ResolutionImpossible(self._failure_causes)
             else:
                 # Pinning was successful. Push a new state to do another pin.
--- a/src/resolvelib/resolvers.py
+++ b/src/resolvelib/resolvers.py
@@ -131,7 +131,7 @@ class Resolution(object):
         state = State(
             mapping=base.mapping.copy(),
             criteria=base.criteria.copy(),
-            backtrack_causes=base.backtrack_causes.copy(),
+            backtrack_causes=base.backtrack_causes[:],
         )
         self._states.append(state)
 
--- a/src/resolvelib/resolvers.py
+++ b/src/resolvelib/resolvers.py
@@ -113,6 +113,7 @@ class Resolution(object):
         self._p = provider
         self._r = reporter
         self._states = []
+        self._failure_casues = []
 
     @property
     def state(self):
@@ -185,6 +186,7 @@ class Resolution(object):
                 self.state.criteria,
                 operator.attrgetter("information"),
             ),
+            failure_causes=self._failure_causes
         )
 
     def _is_current_pin_satisfying(self, name, criterion):
@@ -369,11 +371,12 @@ class Resolution(object):
                 # Backtrack if pinning fails. The backtrack process puts us in
                 # an unpinned state, so we can work on it in the next round.
                 success = self._backtrack()
+                self._failure_causes = [i for c in failure_causes for i in c.information]
 
                 # Dead ends everywhere. Give up.
                 if not success:
                     causes = [i for c in failure_causes for i in c.information]
-                    raise ResolutionImpossible(causes)
+                    raise ResolutionImpossible(self._failure_causes)
             else:
                 # Pinning was successful. Push a new state to do another pin.
                 self._push_new_state()
--- a/src/resolvelib/resolvers.py
+++ b/src/resolvelib/resolvers.py
@@ -113,7 +113,7 @@ class Resolution(object):
         self._p = provider
         self._r = reporter
         self._states = []
-        self._failure_casues = []
+        self._failure_causes = []
 
     @property
     def state(self):
--- a/src/resolvelib/resolvers.py
+++ b/src/resolvelib/resolvers.py
@@ -99,7 +99,7 @@ class ResolutionTooDeep(ResolutionError):
 
 
 # Resolution state in a round.
-State = collections.namedtuple("State", "mapping criteria")
+State = collections.namedtuple("State", "mapping criteria backtrack_causes")
 
 
 class Resolution(object):
@@ -113,7 +113,6 @@ class Resolution(object):
         self._p = provider
         self._r = reporter
         self._states = []
-        self._failure_causes = []
 
     @property
     def state(self):
@@ -132,6 +131,7 @@ class Resolution(object):
         state = State(
             mapping=base.mapping.copy(),
             criteria=base.criteria.copy(),
+            backtrack_causes=base.backtrack_causes.copy(),
         )
         self._states.append(state)
 
@@ -186,7 +186,7 @@ class Resolution(object):
                 self.state.criteria,
                 operator.attrgetter("information"),
             ),
-            failure_causes=self._failure_causes
+            backtrack_causes=self.state.backtrack_causes
         )
 
     def _is_current_pin_satisfying(self, name, criterion):
@@ -234,7 +234,7 @@ class Resolution(object):
             # backtracking looks at this mapping to get the last pin.
             self.state.mapping.pop(name, None)
             self.state.mapping[name] = candidate
-
+    
             return []
 
         # All candidates tried, nothing works. This criterion is a dead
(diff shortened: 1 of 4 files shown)


src/resolvelib/resolvers.py
```diff
@@ -113,6 +113,7 @@ def __init__(self, provider, reporter):
         self._p = provider
         self._r = reporter
         self._states = []
+        self._failure_casues = []
```
bblommers, line 116: This looks like a typo - the other changes use `_failure_causes`

notatallshaw: It's not a typo, not sure what you mean though

pfmoore: This is the only place you use the spelling "casues" (with the "u" before the "e") rather than "causes" ("u" first")...

notatallshaw: Oh I see, yes it's a typo, let me fix that!

src/resolvelib/resolvers.py
```diff
@@ -369,11 +371,12 @@ def resolve(self, requirements, max_rounds):
                 # Backtrack if pinning fails. The backtrack process puts us in
                 # an unpinned state, so we can work on it in the next round.
                 success = self._backtrack()
+                self._failure_causes = [i for c in failure_causes for i in c.information]
 
                 # Dead ends everywhere. Give up.
                 if not success:
                     causes = [i for c in failure_causes for i in c.information]
```
pfmoore, line 378: Can't we drop the (re-) calculation of causes here?

notatallshaw: Fixed. I think this error and the typo crept in when I had to split up my original patch between the different repositories, apologies for that.

src/resolvelib/resolvers.py
```diff
@@ -369,11 +371,11 @@ def resolve(self, requirements, max_rounds):
                 # Backtrack if pinning fails. The backtrack process puts us in
                 # an unpinned state, so we can work on it in the next round.
                 success = self._backtrack()
+                self._failure_causes = [i for c in failure_causes for i in c.information]
```
pradyunsg, line 374: This line is too long, and flake8 is not happy about that.

pradyunsg (changes requested):
This is smart and I like it.

src/resolvelib/resolvers.py
notatallshaw: Also I'm suddenly unsure why I placed this _after_ backtracking. I'm going to experiment a little bit with the placement of this line and also how to make it nicely fit the 80 character limit (which I didn't realize sorry).

pradyunsg: It's all good. :)

FWIW, you can just change this to:

```suggestion
                self._failure_causes = [
                    i for c in failure_causes for i in c.information
                ]
```

notatallshaw: Sorry I accidentally deleted my previous comment. Thanks for the suggestion.

I'm going to leave it broken for just a few hours while I decide where to put the line. I realized there are 3 places it can go and they all make a difference:

1. Right after `failure_causes = ...`
2. Right after `if failure_causes`
3. Where it currently is

And they would mean

1. Limits the change in preference to only when failures are happening in the backtracking
2. The change in preference happens immediately once backtracking starts and the preference would only be changed when a new failure happened
3. Same as 2 but the preference would take effect after backtracking 

I've realized that number 3 is definitely wrong, and that it should be 1 or 2. I'm going to go back and test all my reproducible examples, post the finding to https://github.com/pypa/pip/issues/10479 with the requirements.txts attached, and then update this PR.

uranusjr:
Saving the failure cases directly on the resolver feels wrong to me, since the cases are only specific to a backtracking session and not the entire resolution. I feel it makes more sense to put it on the State objects instead.

notatallshaw:
> Saving the failure causes directly on the resolver feels wrong to me, since the cases are only specific to a backtracking session and not the entire resolution. I feel it makes more sense to put them on the State objects instead (or another stack that only manages backtracking, but I want to avoid that if possible).

I'd not thought about this approach. Taking a quick look at the code pushing new state seems to happen Pip doesn't have new failure causes, so I'm not quite sure how this would be achieved. I'll try and study the code more carefully and think on it.

uranusjr:
Expanding on the thought, `Criterion.incompatibilities` is calculated from similar sources to `failure_causes`, so perhaps we should try to combine the two instead and pass whatever is exposed on `Criterion` to `Provider.get_preferences` instead. This would also make the signature somehow lines up better with `Provider.find_matches`; them being different has always felt weird to me.

uranusjr:
Some more context on why I feel the attribute should be on `State` (either directly or on criterion objects on it) that may help putting things into context.

A resolver is fundamentally traversing deeper and deeper into a graph to find a working solution. This can be most naively implemented by recursion, but you know, Python does not do that particularly well, so the resolver is, at the highest level, unrolling the recursion into a loop, and `states` is the stack that holds context for each iteration that’s done by call frames in recursion. One recursion is one interation in the main loop, and its call frame is created by pushing a `State` object.

With recuision, the best way to implement backtracking is probably with exceptions and re-raising, which bubbles up through the call chain until we find a working call site. Mapping this to iteration, we’d keep popping states, modifying the previous state a bit (the `except` block), and if that still does not work, pop one more frame (re-raise). Which is why I feel putting `failure_causes` on a state feels more natural; the approach you’re taking right now is like introducing global state to functions (`self._failure_causes` is not managed by any call frame and is overwritten on every assignment), which works, but should be avoidable for what we’re trying to achieve, and should be avoided if practical.

notatallshaw:
> Which is why I feel putting `failure_causes` on a state feels more natural; the approach you’re taking right now is like introducing global state to functions (`self._failure_causes` is not managed by any call frame and is overwritten on every assignment), which works, but should be avoidable for what we’re trying to achieve, and should be avoided if practical.

This was very helpful thanks! 

Any solution that involved pinning the failure causes to the state would end up having the same effect as option 1 as I mentioned here: https://github.com/sarugaku/resolvelib/pull/84#discussion_r712414959 . So first I am going to test if this approach still works in the test cases I have.

It seems also that the simplest way to do this is to have `Resolution._attempt_to_pin_criterion` instead of returning causes pin thems to `Resolution.state`, and have `State` have a new property "failure_causes"? Would this be a good approach to attempt or are there some mechanics of `resolvelib` elsewhere that I need to understand / modify instead?

notatallshaw:
> Yeah this would be the easiest approach, but maybe we can redesign `Criterion.incompatibilities` for this. This is a list of known incompatible candidates, I’m wondering we can turn it into a list of kind of wrapping with more information (similar to how `Criterion.information` is a list of wrappers to the criterion’s “parent” candidates) to achieve the same purpose with less duplication.

Okay, I took a look at this and I need a better understanding of how Criterion.incompatibilities work, I'll do some further research.

In the mean time I was very easily able to move failure_causes off being a global variable and on to the state, this slightly changes the behavior of when the failure_causes preferences are applied so I am running over the requirements files I have now to make sure it still solves excessive backtracking. I'll update the PR once done.

After that, I think, migrating to using Criterion.incompatibilities should just be a refactoring and not affect the behavior of backtracking.

src/resolvelib/resolvers.py
notatallshaw: This is now resolved

notatallshaw:
I've not updated the causes to be stored on the state and I am calling them "backtrack_causes" instead of "failure_causes". This is because they are not actually the failures of the current state, but instead failures of a state which needed to be backtracked out of.

I personally feel it makes everything read better, especially in the Provider, But let me know if anyone was particularly tied to "failure_causes"

notatallshaw:
FYI, I consider this PR ready, if maintainers agree I'll start fixing tests.

I took a look at merging the logic with incompatibilities but I think there's a timing issue. The failure causes are generated from a state which is deleted by backtracking. Incompatibilities are generated from the previous state.

I have to save the failure causes before it gets pinned and them to the latest state that exists after backtracking.

But let me know if you're still unhappy with this approach.

uranusjr:
Please fix the tests then.

notatallshaw:
> 
> 
> Please fix the tests then. You can do this in a separate commit, so if we want to switch direction the tests can easily be un-changed with `git revert`.

I pushed a commit that fixes the tests, turned out to be fairly trivial fix.

But  the tests fail on Python 2.7 because `list` does not have a copy method. This is an easy fix but should I fix it? Does `resolvelib` support Python 2.7 still?

uranusjr:
Yes 2.7 is still supported. You can use `foo[:]` instead, this is compatible to all supported Python versions.

notatallshaw:
> Yes 2.7 is still supported. You can use `foo[:]` instead, this is compatible to all supported Python versions.

Yup, already submitted a commit with that and and all tests pass.

src/resolvelib/resolvers.py
```diff
@@ -369,11 +377,13 @@ def resolve(self, requirements, max_rounds):
                 # Backtrack if pinning fails. The backtrack process puts us in
                 # an unpinned state, so we can work on it in the next round.
                 success = self._backtrack()
+                self.state.backtrack_causes[:] = [
+                    i for c in failure_causes for i in c.information
+                ]
```
uranusjr, line 382: Hmm, since `failure_causes` is a list of criteria, maybe `state.backtrack_cause` should be criteria instead? Just nitpicking though, don't need to change it if you don't think it makes sense.

uranusjr (approved):

src/resolvelib/resolvers.py
notatallshaw: I have no opinion, I only wrote it this way as it was the minimal possible change to make this work. So this would be `self.state.backtrack_causes[:] = failure_causes`?

uranusjr: On second though, let's not bother now. This is all internal so we can refactor at any time.

pradyunsg (commented):
LGTM.

I do think we should add tests for the new behaviour though, to ensure that we’re passing valid things with a valid structure in the new argument.

notatallshaw:
> LGTM.
> 
> I do think we should add tests for the new behaviour though, to ensure that we’re passing valid things with a valid structure in the new argument.

I took a look at the existing tests yesterday and I couldn't find anything that tests any of the behavior here. This PR doesn't actually change the behavior of `resolvelib` it just gives more information for downstream library so it in turn can choose to change the ordering of the resolution.

The main breaking change here is passing a new argument to `Resolution._p._get_preference`, but there don't seem to be any tests on what `Resolution._p` is or has to provide?

It also adds an extra property to `State` which is a named tuple, so if any code iterated over `State` instead of accessed it via property names that would break. But as best as I can tell `State` is used only internally to `resolvelib` and it is never iterated?

pradyunsg:
> it just gives more information for downstream library so it in turn can choose to change the ordering of the resolution.

Yup yup, this is what I'm saying we should add a test for.

None the less, I also thought we had more extensive tests for this library than we actually do, so I'm happy to kick the can down the road for this. :)

pradyunsg (approved):

pradyunsg:
Hurrah. @uranusjr do you wanna cut the release, or do you think it might make sense for me to learn how to do that on this project? :)

notatallshaw:
> 
> 
> > it just gives more information for downstream library so it in turn can choose to change the ordering of the resolution.
> 
> Yup yup, this is what I'm saying we should add a test for.
> 
> None the less, I also thought we had more extensive tests for this library than we actually do, so I'm happy to kick the can down the road for this. :)

I had a think about this and thought maybe provider should provide a Base or Abstract class so this can be checked / tested.

Then I looked at the code and realized there is an `AbstractProvider` and this method signature needs to be updated otherwise when Pip vendors this tests will break on Pips side: https://github.com/sarugaku/resolvelib/blob/master/src/resolvelib/providers.py#L12

How do I proceed fixing this PR post-merge? Make a new PR?

notatallshaw:
Created new Pull Request https://github.com/sarugaku/resolvelib/pull/86 due to missing `AbstractProvider`.

uranusjr:
@pradyunsg Feel free try your hands, I can always take over if things go south. There's a tox task to automate release, but to be honest I never remember its exact arguments and always ended up reading the source and manually invoked all the steps instead, it's actually easier than worrying about whether automation is doing the right thing.

notatallshaw:
> 
> 
> @pradyunsg Feel free try your hands, I can always take over if things go south. There's a tox task to automate release, but to be honest I never remember its exact arguments and always ended up reading the source and manually invoked all the steps instead, it's actually easier than worrying about whether automation is doing the right thing.

What are people's thoughts about making it for the pip 21.3 release? Obviously I would of preferred that, but if it can't it can't.

pfmoore:
That's a nice clean release process 🙂 Well done to whoever implemented the automation around that!

uranusjr:
I added the bullets to DEVELOPMENT.rst, thanks for verifying the process works :p

PR #85, opened 2021-09-20
Add a Developing Doc

jbylund:
I'm ashamed to say I do all my docs in github markdown and this is like greek to me.  Also let me know if this should just be inlined into the main readme and if there's anything to add/remove.

### DIFF
--- a/DEVELOPMENT.rst
+++ b/DEVELOPMENT.rst
@@ -2,7 +2,7 @@
 Development
 ===========
 
-resolvelib is a volunteer maintained open source project and we welcome contributions of all forms.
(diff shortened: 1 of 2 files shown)


DEVELOPMENT.rst
```diff
@@ -0,0 +1,27 @@
+===========
+Development
+===========
+
+resolvelib is a volunteer maintained open source project and we welcome contributions of all forms.
```
uranusjr, line 5: Please unify the casing to ResolveLib.

uranusjr (changes requested):

uranusjr (approved):

PR #93, opened 2021-10-30
Implement dedicated causes object

notatallshaw:
This is proof of concept to address issues raised in https://github.com/sarugaku/resolvelib/pull/92#issuecomment-955139430 

Derived information can then be calculated and cached as needed.  In particular the downstream provider can run `identifier in backtrack_causes.names` where `names` is a property that calculates the underlying information lazily and only has to be re-calculated if the underlying `causes` are changed.

I'm trying to see if this approach would have consensuses, several things in the code would need fixing before merging.

### DIFF
--- a/src/resolvelib/resolvers.py
+++ b/src/resolvelib/resolvers.py
@@ -86,6 +86,9 @@ class Causes(object):
     def __eq__(self, other):
         return self.causes == other.causes
 
+    def __bool__(self):
+        return bool(self.causes)
+
 
 class Criterion(object):
     """Representation of possible resolution results of a package.
--- a/src/resolvelib/resolvers.py
+++ b/src/resolvelib/resolvers.py
@@ -88,6 +88,9 @@ class Causes(object):
     def __bool__(self):
         return bool(self.causes)
 
+    def __nonzero__(self):
+        return bool(self.causes)
+
 
 class Criterion(object):
     """Representation of possible resolution results of a package.
--- a/src/resolvelib/resolvers.py
+++ b/src/resolvelib/resolvers.py
@@ -77,7 +77,7 @@ class Causes(object):
         return self._information
 
     def _causes_to_names(self):
-        for c in self.causes:
+        for c in self.information:
             yield c.requirement.name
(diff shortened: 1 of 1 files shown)


notatallshaw:
A nice thing about this approach is it allows the downstream library to be decoupled from the internal structure of the causes, which if someone wanted to significantly change the layout of the objects this would allow without any changes to the downstream library.

uranusjr:
I think I like this approach the best, but we need to go a bit further down the decoupling path. Taking things like pypa/pip#10478 into account, perhaps we should take a similar approach to Requirement and Candidate and add an interface in Provider custom Cause object logic? (We can provide a default implementation if that’s a good idea.)

hswong3i:
So #97 and 1.0.0 should now depends on this PR?

uranusjr:
Yes

hswong3i:
Seems a huge workload for Cause, which may need more than a month for implement it precisely… Maybe this breaking change could go into 2.0.0? Or just implement the API interface changes now for 1.0.0 without indeed and perfect implementation, yet?

uranusjr:
Yes except I want it to be a provider method and create the entire cause object, not modify it.

notatallshaw:
Not had any time to work on this this last month. However I expect to have some free time during the holidays. I'm probably going to create a new PR and close this one, implementing it like Requirement and Candidate the way @uranusjr suggests.

I don't feel any ownership over doing the work 😉, so if anyone else wants to take the approach and implement it themselves feel free to, otherwise I will come back with an update later this month.

notatallshaw:
> I think I like this approach the best, but we need to go a bit further down the decoupling path. Taking things like [pypa/pip#10478](https://github.com/pypa/pip/issues/10478) into account, perhaps we should take a similar approach to Requirement and Candidate and add an interface in Provider custom Cause object logic? (We can provide a default implementation if that’s a good idea.)

I've addressed these comments in a new PR: https://github.com/sarugaku/resolvelib/pull/99 and an accompanying pip PR: https://github.com/pypa/pip/pull/10732

PR #94, opened 2021-10-31
Fix mypy failures

notatallshaw:
Related to https://github.com/python/mypy/issues/1153

This became an issue with mypy running on Python 3.10 (I guess `collections.Mapping` and `collections.Sequence` were finally removed).

It's causing mypy to fail in latest PRs: https://github.com/sarugaku/resolvelib/pull/93

### DIFF
--- a/src/resolvelib/compat/collections_abc.py
+++ b/src/resolvelib/compat/collections_abc.py
@@ -3,4 +3,4 @@ __all__ = ["Mapping", "Sequence"]
 try:
     from collections.abc import Mapping, Sequence
 except ImportError:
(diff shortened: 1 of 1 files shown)


uranusjr:
Adding a `compat.pyi` file is probably more consistent to how typing is handled in the repo.

notatallshaw:
> 
> 
> Adding a `compat.pyi` file is probably more consistent to how typing is handled in the repo.

Oh yes, I missed because I forget about `pyi` files. Created a new PR: https://github.com/sarugaku/resolvelib/pull/95

PR #95, opened 2021-10-31
Fix mypy on Python 3.10

notatallshaw:
As collections.Mapping and collections.Sequence have been removed, fixes CI workflow as it uses Python 3.10.

Using `pyi` as per https://github.com/sarugaku/resolvelib/pull/94#issuecomment-955782467


src/resolvelib/compat/collections_abc.pyi
```diff
@@ -0,0 +1 @@
+from collections.abc import Mapping, Sequence
```
uranusjr, line 1:
```suggestion
from collections.abc import Mapping, Sequence

```

PR #96, opened 2021-11-01
Consumer passes in a prepare method which is called over backtrack causes

jbylund:
Add a prepare attribute to the Resolution object which is called on backtrack causes before passing that to `_get_preference`.  This allows providers to operate over a processed version of backtrack_causes without having to implement a cache consumer side (which can be "yucky").

### DIFF
--- a/src/resolvelib/resolvers.py
+++ b/src/resolvelib/resolvers.py
@@ -108,11 +108,10 @@ class Resolution(object):
     the resolution process, and holds the results afterwards.
     """
 
-    def __init__(self, provider, reporter, prepare_causes=None):
+    def __init__(self, provider, reporter):
         self._p = provider
         self._r = reporter
         self._states = []
-        self.prepare = prepare_causes
 
     @property
     def state(self):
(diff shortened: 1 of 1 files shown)


jbylund:
@notatallshaw this is the other possibility I mentioned. Let me know if you think this would mesh well with your ideas for backtracking in the future?

notatallshaw:
This approach LGTM if maintainers want the logic of processing the causes to remain on the downstream library side.

And I continue to really prefer avoiding a cache that depends on the ids of mutable objects (even if theyare carefully used).

I guess let's just see what maintainers think.

uranusjr:
I think #93 would supercede this now?

jbylund:
> I think #93 would supercede this now?

My opinion is that this option would allow for:
* the possibility of a simpler solution making use of stdlib collection types - class is reduced to a function which returns a set
* more separation/flexibility between different consumers of resolvelib - users that might care about the order of conflicts (I don't even know if that could theoretically matter to someone) could choose the passthrough method

uranusjr:
Sounds like a provider function is also a fit for those reasons? If the consumer wishes, it can simply implement the function to return a set, or whatever built-in or custom type that suits the need.

jbylund:
> Sounds like a provider function is also a fit for those reasons? If the consumer wishes, it can simply implement the function to return a set, or whatever built-in or custom type that suits the need.

But I think we need this "hook" to allow the provider to inject such a function?

jbylund:
The pip side of this change would be [this](https://github.com/pypa/pip/compare/main...jbylund:joe/use_prepare_on_provider).

notatallshaw:
I beleive this was effectively implemented by https://github.com/sarugaku/resolvelib/pull/145

And the PR to take advantage of it on the pip side is https://github.com/pypa/pip/pull/13253

Let me know if I missed something.

PR #99, opened 2021-12-18
Let downstream library to define causes object

notatallshaw:
Addresses comments made in https://github.com/sarugaku/resolvelib/pull/93#issuecomment-970470701

Provider now has a `causes` method that allows it to define what the causes object looks like.

The returned object must implement the following:
 * Have truthiness
 * Be copyiable using `copy.copy`

For example if would be possible to implement the method like so to preserve the existing behavior (would work on Python 2.7 to 3.10):

```
def causes(self, causes):
    return [i for c in causes for i in c.information]
```

There is an accompanying pip PR (https://github.com/pypa/pip/pull/10732) that leverages this new interface to resolve the performance issues described in https://github.com/pypa/pip/pull/10621

Need some hints / help defining a better type than "Any" for "causes" if that matters.

### DIFF
--- a/examples/extras_provider.py
+++ b/examples/extras_provider.py
@@ -52,3 +52,6 @@ class ExtrasProvider(AbstractProvider):
             req = self.get_base_requirement(candidate)
             deps.append(req)
         return deps
+
+    def causes(self, causes):
+        return [i for c in causes for i in c.information]
--- a/src/resolvelib/providers.pyi
+++ b/src/resolvelib/providers.pyi
@@ -1,14 +1,4 @@
-from typing import (
-    Any,
-    Collection,
-    Generic,
-    Iterable,
-    Iterator,
-    Mapping,
-    Optional,
-    Protocol,
-    Union,
-)
+from typing import Any, Generic, Iterable, Iterator, Mapping, Protocol, Union
 
 from .reporters import BaseReporter
 from .resolvers import RequirementInformation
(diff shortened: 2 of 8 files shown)


notatallshaw:
There are some architectural differences between the way I've implemented this and how `Requirement` and `Candidate`.

@uranusjr if you get a chance let me know if this is what you were thinking or if I missed the shot and need to make some correction.

notatallshaw:
> The returned object must implement the following:
> 
>     * Have truthiness
>     * Be copyiable using `copy.copy`
> 

FWIW I tried creating a "Protocol" for the copy part of this but I could not get it to pass MyPy when I ran it in Pip. I tried making a protocol, type hinting an argument on a method of the Abstract class and then in Pip type hinting the concrete class with a type that implements that protocol all I ever got was the error: 

```
Argument 5 of
"get_preference" is incompatible with supertype "AbstractProvider"; supertype defines the argument type as "Copyable"  [override]
        def get_preference(
        ^
src\pip\_internal\resolution\resolvelib\provider.py:109: note: This violates the Liskov substitution principle
src\pip\_internal\resolution\resolvelib\provider.py:109: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
```

I spent an hour or so trying to read / understand what I was missing but I couldn't figure it out. If anyone has any hints / docs / tutorials that I could read / watch I'd be appreciating.

Even if I got the copy part to work I don't know if I can even create a "Protocol" for the truthiness part of this because resolvelib supports Python 2.7 still, so the protocol would need to express that it should implement `__nonzero__` for Python 2.7  and `__bool__` for everything else. Seems less of an issue though.

notatallshaw:
Upon some further thinking this approach is type unsafe, I need to study how other objects are implemented a bit better first.

PR #105, opened 2022-05-06
resolvelib/providers.pyi: add hint for `backtrack_causes`

### LINKED ISSUE #104: Typing inconsistency: `AbstractProvider.get_preferences`
Here's how `AbstractProvider.get_preferences` is defined:

```py
    def get_preference(
        self,
        identifier,
        resolutions,
        candidates,
        information,
        backtrack_causes,
    ):
```

and here's the `.pyi` definition:

```py
    def get_preference(
        self,
        identifier: KT,
        resolutions: Mapping[KT, CT],
        candidates: Mapping[KT, Iterator[CT]],
        information: Mapping[KT, Iterator[RequirementInformation[RT, CT]]],
    ) -> Preference: ...
```

Note that the latter is missing `backtrack_causes`, so any downstream consumers of `resolvelib`'s type hints won't be able to use the signature. 

I'm happy to make a small PR fixing this!

woodruffw:
This is based on my reading of the docs:

```
        :param backtrack_causes: Sequence of requirement information that were
            the requirements that caused the resolver to most recently backtrack.
```

Closes #104.

Signed-off-by: William Woodruff <william@trailofbits.com>

### DIFF
--- a/src/resolvelib/providers.pyi
+++ b/src/resolvelib/providers.pyi
@@ -1,12 +1,11 @@
 from typing import (
     Any,
-    Collection,
     Generic,
     Iterable,
     Iterator,
     Mapping,
-    Optional,
     Protocol,
+    Sequence,
     Union,
 )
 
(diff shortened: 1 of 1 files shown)


pradyunsg:
Thanks for catching this inconsistency @woodruffw!

woodruffw:
Thanks for the merge! No rush at all, but a patch release so that I could update `pip-audit`'s typing would be greatly appreciated 🙂

PR #107, opened 2022-05-27
Lazy sequence iterview

frostming:
Result iterable returned by `find_matches()` won't be consumed all at once.
Also unite the `_FactoryIterableView` with `_SequenceIterableView`
With the change, the underlying iterable is guaranteed to be consumed only once.

### DIFF
--- a/src/resolvelib/structs.py
+++ b/src/resolvelib/structs.py
@@ -116,7 +116,8 @@ class _FactoryIterableView(object):
     """
 
     def __init__(self, factory):
-        self._iterable = factory()
+        self._factory = factory
+        self._iterable = None
 
     def __repr__(self):
         return "{}({})".format(type(self).__name__, list(self))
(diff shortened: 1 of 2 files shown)


uranusjr:
I intentionally separated the two wrapper implementations to help debugging. `_SequenceIterableView` holds the original sequence, which can be random-accessed and iterated over multiple times, which is useful in the debugger to better identify the situation.

frostming:
@uranusjr Fine, I revert the change to keep the `SequenceIterableView`, while `FactoryIterableView` is still optimized to avoid calling `factory` multiple times, especially when retrieving the first value is expensive.

frostming:
@uranusjr Can you revisit this PR? I think it is worth merging since it will save a lot if it is expensive to construct the first candidate.

src/resolvelib/structs.py
```diff
@@ -116,22 +116,23 @@ class _FactoryIterableView(object):
     """
 
     def __init__(self, factory):
-        self._factory = factory
+        self._iterable = factory()
```
uranusjr, line 119: Is it possible to not call the factory here? It would be nice to not needing to execute the code if the iterable is never accessed.

uranusjr (approved):
One minor nit

PR #111, opened 2022-10-12
Fix reject of valid state in criteria compatibility check

### LINKED ISSUE #91: criteria compatibility check while backtracking includes dependencies for version we're backtracking on
# Context
While using pip, which makes use of this library, I noticed that `pip install flake8 flake8-isort` resulted in backtracking over all flake8's versions, not finding a suitable candidate, followed by backtracking over flake8-isort's versions until a suitable (but very old) candidate was found there. When I looked at these projects I didn't see any reason why none of the flake8 versions would be compatible with the latest flake8-isort. Indeed, reversing the order (`pip install flake8-isort`) installs the latest flake8-isort and a compatible flake8 as expected, only having to backtrack once on flake8.
When I noticed this seemingly inconsistent behavior I added some breakpoints to this library's code and tried to get a grasp of what was going on. I should note as a disclaimer that I haven't worked with this code before, so I might just be missing something.

# Concrete
Here's a timeline of what I believe happens for `pip install flake8 flake8-isort`:
1. The latest flake8 (`4.0.1` at the time of writing) gets selected.
2. The latest flake8-isort gets selected
3. The flake8-isort candidate requires `flake8<4`, which isn't compatible with the flake8 candidate, so we backtrack on flake8.
4. For each flake8 version, we check if its dependencies are compatible with the current criteria: https://github.com/sarugaku/resolvelib/blob/62c56b587c69a42c8f006e032bbf17a63df54d45/src/resolvelib/resolvers.py#L203-L204
5. This fails with
```
> /home/sander/.virtualenvs/temp/lib/python3.9/site-packages/pip/_vendor/resolvelib/resolvers.py(218)_attempt_to_pin_criterion()
-> continue
(Pdb) l
213  	            try:
214  	                criteria = self._get_updated_criteria(candidate)
215  	            except RequirementsConflicted as e:
216  	                causes.append(e.criterion)
217  	                breakpoint()
218  ->	                continue
219  	
220  	            # Check the newly-pinned candidate actually works. This should
221  	            # always pass under normal circumstances, but in the case of a
222  	            # faulty provider, we will raise an error to notify the implementer
223  	            # to fix find_matches() and/or is_satisfied_by().
(Pdb) e.criterion
Criterion((SpecifierRequirement('pyflakes<2.5.0,>=2.4.0'), via=LinkCandidate('https://files.pythonhosted.org/packages/34/39/cde2c8a227abb4f9ce62fe55586b920f438f1d2903a1a22514d0b982c333/flake8-4.0.1-py2.py3-none-any.whl#sha256=479b1304f72536a55948cb40a32dce8bb0ffe3501e26eaf292c7e60eb5e0428d (from https://pypi.org/simple/flake8/) (requires-python:>=3.6)')), (SpecifierRequirement('pyflakes<2.4.0,>=2.3.0'), via=LinkCandidate('https://files.pythonhosted.org/packages/fc/80/35a0716e5d5101e643404dabd20f07f5528a21f3ef4032d31a49c913237b/flake8-3.9.2-py2.py3-none-any.whl#sha256=bf8fd333346d844f616e8d47905ef3a3384edae6b4e9beb0c5101e25e3110907 (from https://pypi.org/simple/flake8/) (requires-python:!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*,>=2.7)')))
```

This seems to indicate that the dependency compatibility check includes the dependencies of the version we're backtracking on. Since the currently selected flake8 has a dependency constraint which is incompatible with the constraints for all older versions, backtracking fails. In practice, the latest `flake8<4` is compatible with the rest of the current criteria.

sanderr:
Fix incorrect rejection of some states where dependencies for multiple versions of the same package are considered for conflicts. This pull request introduces a check at the end of each round of pinning: any newly invalidated pin (because of a new requirement), is removed as an information provider from the state's criteria.

I believe this implements the same behavior as initially proposed in #91 but more directed, to maintain the overall correctness of each state's criteria and to limit overhead.

I am currently still unsure how I should address [this comment](https://github.com/sarugaku/resolvelib/issues/91#issuecomment-950679928) (last remaining TODO in the diff) but I thought I'd already open the bulk of my changes for review.

(If accepted, might I ask to add the [hacktoberfest-accepted](https://hacktoberfest.com/participation/) label to the pull request? I mean to follow through on this either way of course.)

closes #91


src/resolvelib/resolvers.py
```diff
@@ -173,6 +173,33 @@ def _add_to_criteria(self, criteria, requirement, parent):
...
+    def _remove_information_from_criteria(self, criteria, parents):
+        """
+        Removes information from a set of parents from a criteria. Concretely, removes
+        from each criterion's `information` field all values that have one of `parents`
+        as provider of the requirement.
+
+        :param criteria: The criteria to update.
+        :param parents: The set of identifiers for which to remove information from all
+            criteria.
+        """
```
uranusjr, line 185:
```suggestion
        """Remove information from parents of criteria.
        
        Concretely, removes all values from each criterion's ``information``
        field that have one of ``parents`` as provider of the requirement.

        :param criteria: The criteria to update.
        :param parents: Identifiers for which to remove information from all criteria.
        """
```

src/resolvelib/resolvers.py
```diff
@@ -173,6 +173,33 @@ def _add_to_criteria(self, criteria, requirement, parent):
...
+        if not parents:
+            return
+        for key, criterion in criteria.items():
+            criteria[key] = Criterion(
+                criterion.candidates,
+                # TODO: is empty information allowed?
```
uranusjr, line 191: Empty information is technically _allowed_ but likely unexpected, since it records where a criterion came from, and if a criterion comes from nowhere, should it not be removed?

src/resolvelib/resolvers.py
```diff
@@ -383,6 +414,17 @@ def resolve(self, requirements, max_rounds):
                 if not success:
                     raise ResolutionImpossible(self.state.backtrack_causes)
             else:
+                # discard as information sources any invalidated names
+                # (unsatisfied names that were previously satisfied)
+                newly_unsatisfied_names = {
+                    key
+                    for key, criterion in self.state.criteria.items()
+                    if key in satisfied_names
+                    if not self._is_current_pin_satisfying(key, criterion)
```
frostming, line 423:
```suggestion
                    and not self._is_current_pin_satisfying(key, criterion)
```
Maybe you mean this?

src/resolvelib/resolvers.py
sanderr: My reasoning was that we're dropping the constraint provided by a parent because we're dropping the pin on that parent. But if the new pin on the parent happens to have the same constraint (the common case), we might want to keep `incompatibilities`, in order to prevent backtracking over the same thing again.
But as I type this I realize that I'm really not yet confident in my understanding of how incompatibilities are gathered. So if you are confident that we can safely drop the criterion when no information remains I'll apply that change. Otherwise I'll have to sink my teeth in the incompatibilities some more.

src/resolvelib/resolvers.py
sanderr: They're equivalent actually.

frostming:
Hi, I've accepted this PR for hacktoberfest, can you fix the lint issue by formatting the code with black?

src/resolvelib/resolvers.py
uranusjr: If the intent is to keep `incompatibilities` in the graph, I think it’s OK to keep a criterion with empty information (it’s basically an orphan node). We should describe this in the release note (by adding a release note fragment file in `news`) so provider/reporter implementers are aware of the possibility that `information` _could_ be empty (and what that means), it’s likely no-one would be particular troubled by this since an orphan node can simply be ignored in the final graph anyway.

sanderr: All right, I'll do that. I think `pip` will need a minor change, but there seems to already be a breaking change in master compared to the last release anyway so I guess that is acceptable.

frostming: Does the pin need to be excluded from the final result whose criterion has empty information? Since `_is_current_pin_satisfying()` will always return True if there has been one pinned before.

frostming: I think the orphan nodes will be excluded automatically by `build_result()`

sanderr: I've made an attempt at the news fragment but I'm not entirely sure about your change entry conventions.

tests/test_resolvers.py
```diff
@@ -1,4 +1,18 @@
...
+    Mapping,
+    Sequence,
+    Set,
+    Tuple,
+    Union,
+)
+
 import pytest
+from packaging.version import Version
+from pkg_resources import Requirement
```
uranusjr, line 15: Can this use `packaging.requirements.Requirement` instead? I would very much prefer to not depend on setuptools.

tests/test_resolvers.py
```diff
@@ -7,6 +21,12 @@
     ResolutionImpossible,
     Resolver,
 )
+from resolvelib.resolvers import Resolution  # type: ignore
```
uranusjr, line 24: What errors does this ignore?

tests/test_resolvers.py
```diff
@@ -12,7 +12,7 @@
 
 import pytest
 from packaging.version import Version
-from pkg_resources import Requirement
+from packaging.requirements import Requirement
```
uranusjr, line 15: Sort this alphabetically please

tests/test_resolvers.py
sanderr: `Resolution` is not part of the public interface declared in `src/resolvelib/resolvers.pyi`. My reasoning was that for testing that does not matter. I meant to ask about this but I forgot about it. The error is as follows:
```
tests/test_resolvers.py:24: error: Module "resolvelib.resolvers" has no attribute "Resolution"; maybe "ResolutionError"?
Found 1 error in 1 file (checked 1 source file)
```

uranusjr: We can add `Resolution` to the type stubs; I think someone would need it at some point anyway.

tests/test_resolvers.py
sanderr: Done

tests/test_resolvers.py
sanderr: Done

tests/test_resolvers.py
sanderr: All right, I'll do that, thanks.

sanderr: Done. I did have to add another more specific type ignore because the test case monkeypatches a private method. I think that should be acceptable?

src/resolvelib/resolvers.pyi
```diff
@@ -55,6 +55,18 @@ class ResolutionImpossible(ResolutionError, Generic[RT, CT]):
 class ResolutionTooDeep(ResolutionError):
     round_count: int
 
+# This should be a NamedTuple, but Python 3.6 has a bug that prevents it.
+# https://stackoverflow.com/a/50531189/1376863
```
sanderr, line 59: I copied this from another tuple at the top of the file.

uranusjr (approved):

Issue #103, opened 2022-05-06
0.8.1: pytest warnings


kloczek:
I'm trying to package your module as an rpm package. So I'm using the typical PEP517 based build, install and test cycle used on building packages from non-root account.
- `python3 -sBm build -w --no-isolation`
- because I'm calling `build` with `--no-isolation` I'm using during all processes only locally installed modules
- install .whl file in </install/prefix>
- run pytest with PYTHONPATH pointing to sitearch and sitelib inside </install/prefix>

Here is pytest output:
```console
+ PYTHONPATH=/home/tkloczko/rpmbuild/BUILDROOT/python-resolvelib-0.8.1-2.fc35.x86_64/usr/lib64/python3.8/site-packages:/home/tkloczko/rpmbuild/BUILDROOT/python-resolvelib-0.8.1-2.fc35.x86_64/usr/lib/python3.8/site-packages
+ /usr/bin/pytest -ra
=========================================================================== test session starts ============================================================================
platform linux -- Python 3.8.13, pytest-7.1.2, pluggy-1.0.0
rootdir: /home/tkloczko/rpmbuild/BUILD/resolvelib-0.8.1
collected 48 items

tests/test_resolvers.py ....                                                                                                                                         [  8%]
tests/test_structs.py .......                                                                                                                                        [ 22%]
tests/functional/cocoapods/test_resolvers_cocoapods.py x........x...............                                                                                     [ 75%]
tests/functional/python/test_resolvers_python.py ....xx..                                                                                                            [ 91%]
tests/functional/swift-package-manager/test_resolvers_swift.py ....                                                                                                  [100%]

============================================================================= warnings summary =============================================================================
tests/functional/cocoapods/test_resolvers_cocoapods.py: 188 warnings
tests/functional/python/test_resolvers_python.py: 20 warnings
  /usr/lib/python3.8/site-packages/packaging/version.py:111: DeprecationWarning: Creating a LegacyVersion has been deprecated and will be removed in the next major release
    warnings.warn(

tests/functional/python/test_resolvers_python.py: 27 warnings
  /usr/lib/python3.8/site-packages/packaging/specifiers.py:255: DeprecationWarning: Creating a LegacyVersion has been deprecated and will be removed in the next major release
    warnings.warn(

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
========================================================================= short test summary info ==========================================================================
[...]

PR #112, opened 2022-10-27
fixed invalid version specifiers in a test index

sanderr:
I believe this resolves part of #103. It removes all `LegacyVersion` deprecation warnings for the Python tests. The others are more tricky because they aren't versions for Python packages, therefore they aren't PEP440 compliant. I did not address those for now.

If accepted (within the month of October, which may be a bit of a stretch at this point), might I ask once more to add the [hacktoberfest-accepted](https://hacktoberfest.com/participation/) label to the pull request?

### DIFF
--- a/tests/functional/python/test_resolvers_python.py
+++ b/tests/functional/python/test_resolvers_python.py
@@ -129,7 +129,6 @@ CASE_NAMES = [name for name in os.listdir(CASE_DIR) if name.endswith(".json")]
 
 XFAIL_CASES = {
     "pyrex-1.9.8.json": "Too many rounds (>500)",
-    "same-package-extras.json": "State not cleaned up correctly",
 }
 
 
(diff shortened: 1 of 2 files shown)


tests/functional/python/test_resolvers_python.py
```diff
@@ -129,7 +129,6 @@ def get_dependencies(self, candidate):
 
 XFAIL_CASES = {
     "pyrex-1.9.8.json": "Too many rounds (>500)",
-    "same-package-extras.json": "State not cleaned up correctly",
```
sanderr, line 132: I am wondering about this. The change above caused this test to fail but I'm afraid I don't understand the reason it was marked XFAIL in the first place. Is my fix even valid?

uranusjr: It could be that the issue was fixed; we did make some improvements around the area a while ago. The test case _should_ pass, so it’s a good idea to remove this xfail if it does.

uranusjr (approved):

PR #113, opened 2022-12-23
Implement backjumping over unrelated states

bennati:
The current backtracking logic assumes that the last state in the stack
is the state that caused the incompatibility. This is not always the case.
For example, given three requirements A, B and C, with dependencies
A1, B1 and C1, where A1 and B1 are incompatible.
The requirements are processed one after the other, so the last state
is related to C, while the incompatibility is caused by B.

The current behavior causes significant slowdowns in case there are
many candidates for B and C, as all their combination are evaluated
before a compatible version of B can be found.

The new behavior discards a state if the packages that cause the
incompatibility are not found among the direct dependencies
of the candidate in the current state.
In our example, this causes the state related to C to be dropped
without evaluating any of its candidates, until a compatible candidate
of B is found.

This change can be tested with the following requirements:
```
boto3==1.10.16
s3fs
seaborn
```

### DIFF
--- a/src/resolvelib/resolvers.py
+++ b/src/resolvelib/resolvers.py
@@ -266,8 +266,8 @@ class Resolution(object):
         # end, signal for backtracking.
         return causes
 
-    def _backtrack(self):
-        """Perform backtracking.
+    def _backjump(self, causes):
+        """Perform backjumping.
 
         When we enter here, the stack is like this::
 
@@ -283,22 +283,44 @@ class Resolution(object):
 
         Each iteration of the loop will:
 
-        1.  Discard Z.
-        2.  Discard Y but remember its incompatibility information gathered
+        1.  Identify Z. The incompatibility is not always caused by the latest state.
+            For example, given three requirements A, B and C, with dependencies
+            A1, B1 and C1, where A1 and B1 are incompatible: the last state
+            might be related to C, so we want to discard the previous state.
+        2.  Discard Z.
+        3.  Discard Y but remember its incompatibility information gathered
             previously, and the failure we're dealing with right now.
-        3.  Push a new state Y' based on X, and apply the incompatibility
+        4.  Push a new state Y' based on X, and apply the incompatibility
             information from Y to Y'.
-        4a. If this causes Y' to conflict, we need to backtrack again. Make Y'
+        5a. If this causes Y' to conflict, we need to backtrack again. Make Y'
             the new Z and go back to step 2.
-        4b. If the incompatibilities apply cleanly, end backtracking.
+        5b. If the incompatibilities apply cleanly, end backtracking.
         """
+        incompatible_deps = set(
+            [c.parent.name for c in causes if c.parent is not None]
+            + [c.requirement.name for c in causes]
+        )
         while len(self._states) >= 3:
             # Remove the state that triggered backtracking.
             del self._states[-1]
 
-            # Retrieve the last candidate pin and known incompatibilities.
-            broken_state = self._states.pop()
-            name, candidate = broken_state.mapping.popitem()
+            # Ensure to backtrack to a state that caused the incompatibility
+            incompatible_state = False
+            while not incompatible_state:
+                # Retrieve the last candidate pin and known incompatibilities.
+                try:
+                    broken_state = self._states.pop()
+                    name, candidate = broken_state.mapping.popitem()
+                except (IndexError, KeyError):
+                    raise ResolutionImpossible(causes)
+                current_dependencies = {
+                    self._p.identify(d)
+                    for d in self._p.get_dependencies(candidate)
+                }
+                incompatible_state = not current_dependencies.isdisjoint(
+                    incompatible_deps
+                )
+
             incompatibilities_from_broken = [
                 (k, list(v.incompatibilities))
                 for k, v in broken_state.criteria.items()
@@ -403,10 +425,10 @@ class Resolution(object):
 
             if failure_causes:
                 causes = [i for c in failure_causes for i in c.information]
-                # Backtrack if pinning fails. The backtrack process puts us in
+                # Backjump if pinning fails. The backjump process puts us in
                 # an unpinned state, so we can work on it in the next round.
                 self._r.resolving_conflicts(causes=causes)
-                success = self._backtrack()
+                success = self._backjump(causes)
                 self.state.backtrack_causes[:] = causes
 
                 # Dead ends everywhere. Give up.
--- a/news/113.feature.rst
+++ b/news/113.feature.rst
@@ -1,0 +1,1 @@
+Implement backjumping to significantly speed up the resolution process by skipping over irrelevant parts of the resolution search space.
--- a/tests/test_resolvers.py
+++ b/tests/test_resolvers.py
@@ -1,14 +1,17 @@
-from typing import (
-    Any,
-    Iterable,
-    Iterator,
-    List,
-    Mapping,
-    Sequence,
-    Set,
-    Tuple,
-    Union,
-)
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from typing import (
+        Any,
+        Iterable,
+        Iterator,
+        List,
+        Mapping,
+        Sequence,
+        Set,
+        Tuple,
+        Union,
+    )
 
 import pytest
 from packaging.requirements import Requirement
@@ -21,12 +24,17 @@ from resolvelib import (
     ResolutionImpossible,
     Resolver,
 )
-from resolvelib.resolvers import (
-    Criterion,
-    RequirementInformation,
-    RequirementsConflicted,
-    Resolution,
-)
+
+if TYPE_CHECKING:
+    from resolvelib.resolvers import (
+        Criterion,
+        RequirementInformation,
+        RequirementsConflicted,
+    )
+
+from collections import namedtuple
+
+from resolvelib.resolvers import Resolution
 
 
 def test_candidate_inconsistent_error():
@@ -115,10 +123,21 @@ def test_candidate_depends_on_requirements_of_same_identifier(specifiers):
 
 
 def test_resolving_conflicts():
+    Candidate = namedtuple(
+        "Candidate", ["name", "version", "requirements"]
+    )  # name, version, requirements
+    _Requirement = namedtuple(
+        "Requirement", ["name", "versions"]
+    )  # name, versions
+    a1 = Candidate("a", 1, [_Requirement("q", {1})])
+    a2 = Candidate("a", 2, [_Requirement("q", {2})])
+    b = Candidate("b", 1, [_Requirement("q", {1})])
+    q1 = Candidate("q", 1, [])
+    q2 = Candidate("q", 2, [])
     all_candidates = {
-        "a": [("a", 1, [("q", {1})]), ("a", 2, [("q", {2})])],
-        "b": [("b", 1, [("q", {1})])],
-        "q": [("q", 1, []), ("q", 2, [])],
+        "a": [a1, a2],
+        "b": [b],
+        "q": [q1, q2],
     }
 
     class Reporter(BaseReporter):
@@ -136,20 +155,22 @@ def test_resolving_conflicts():
             return 0
 
         def get_dependencies(self, candidate):
-            return candidate[2]
+            return candidate.requirements
 
         def find_matches(self, identifier, requirements, incompatibilities):
-            bad_versions = {c[1] for c in incompatibilities[identifier]}
+            bad_versions = {c.version for c in incompatibilities[identifier]}
             candidates = [
                 c
                 for c in all_candidates[identifier]
-                if all(c[1] in r[1] for r in requirements[identifier])
-                and c[1] not in bad_versions
+                if all(
+                    c.version in r.versions for r in requirements[identifier]
+                )
+                and c.version not in bad_versions
             ]
-            return sorted(candidates, key=lambda c: c[1], reverse=True)
+            return sorted(candidates, key=lambda c: c.version, reverse=True)
 
         def is_satisfied_by(self, requirement, candidate):
-            return candidate[1] in requirement[1]
+            return candidate.version in requirement.versions
 
     def run_resolver(*args):
         reporter = Reporter()
@@ -160,8 +181,12 @@ def test_resolving_conflicts():
         except ResolutionImpossible as e:
             return e.causes
 
-    backtracking_causes = run_resolver([("a", {1, 2}), ("b", {1})])
-    exception_causes = run_resolver([("a", {2}), ("b", {1})])
+    backtracking_causes = run_resolver(
+        [_Requirement("a", {1, 2}), _Requirement("b", {1})]
+    )
+    exception_causes = run_resolver(
+        [_Requirement("a", {2}), _Requirement("b", {1})]
+    )
     assert exception_causes == backtracking_causes
 
 
@@ -171,9 +196,10 @@ def test_pin_conflict_with_self(monkeypatch, reporter):
     Verify correct behavior of attempting to pin a candidate version that conflicts
     with a previously pinned (now invalidated) version for that same candidate (#91).
     """
-    Candidate = Tuple[  # noqa: F841
-        str, Version, Sequence[str]
-    ]  # name, version, requirements
+    if TYPE_CHECKING:
+        Candidate = Tuple[  # noqa: F841
+            str, Version, Sequence[str]
+        ]  # name, version, requirements
     all_candidates = {
         "parent": [("parent", Version("1"), ["child<2"])],
         "child": [


pombredanne:
This is an awesome finding!

pombredanne:
@bennati Could you add a test?

src/resolvelib/resolvers.py
```diff
@@ -412,6 +412,13 @@ def resolve(self, requirements, max_rounds):
                 # Dead ends everywhere. Give up.
                 if not success:
                     raise ResolutionImpossible(self.state.backtrack_causes)
+                else:
+                    criterion = self.state.criteria[name]
+                    # Put newly-pinned candidate at the end. This is essential because
+                    # backtracking looks at this mapping to get the last pin.
+                    for candidate in criterion.candidates:
+                        self.state.mapping.pop(name, None)
+                        self.state.mapping[name] = candidate
```
uranusjr, line 421: Since `name` is unchanged throughout this section, this loop essentially popping and setting different objects to the same key over and over. Which seems wrong to me…?

bennati:
I realized the error is something else entirely, please review again :)

pradyunsg:
If the change does what the message is implying, I'd prefer that we rename the function to _backjump :)

bennati:
Why do you want to rename the function? The function is not jumping anywhere, it is backtracking states in order until it reaches the correct state.

pradyunsg:
https://en.m.wikipedia.org/wiki/Back jumping is why I've suggested the name that I suggested. :)

src/resolvelib/resolvers.py
```diff
@@ -283,22 +283,40 @@ def _backtrack(self):
...
-            name, candidate = broken_state.mapping.popitem()
+            # Ensure to backtrack to a state that caused the incompatibility
+            incompatible_state = False
+            while not incompatible_state:
+                # Retrieve the last candidate pin and known incompatibilities.
+                broken_state = self._states.pop()
+                name, candidate = broken_state.mapping.popitem()
+                current_dependencies = set(
+                    [d.name for d in self._p.get_dependencies(candidate)]
+                )
```
frostming, line 311:
```suggestion
                current_dependencies = {
                    self._p.identify(d) for d in self._p.get_dependencies(candidate)
                }
```

src/resolvelib/resolvers.py
```diff
@@ -283,22 +283,40 @@ def _backtrack(self):
...
+                incompatible_deps = set(
+                    [c.parent.name for c in causes]
+                )  # Parent should never be null
```
frostming, line 314:
```suggestion
                incompatible_deps = {
                    self._p.identify(c.parent) for c in causes
                }  # Parent should never be null
```

src/resolvelib/resolvers.py
```diff
@@ -283,22 +283,40 @@ def _backtrack(self):
...
+                incompatible_state = (
+                    current_dependencies.intersection(incompatible_deps)
+                    != set()
```
frostming, line 317:
```suggestion
                    not current_dependencies.isdisjoint(incompatible_deps)
```

src/resolvelib/resolvers.py
```diff
@@ -283,22 +283,40 @@ def _backtrack(self):
...
         while len(self._states) >= 3:
             # Remove the state that triggered backtracking.
             del self._states[-1]
 
-            # Retrieve the last candidate pin and known incompatibilities.
-            broken_state = self._states.pop()
-            name, candidate = broken_state.mapping.popitem()
+            # Ensure to backtrack to a state that caused the incompatibility
+            incompatible_state = False
+            while not incompatible_state:
```
frostming, line 305: Is it sure it can find the incompatible_state before the states are exhausted?

frostming (changes requested):

src/resolvelib/resolvers.py
uranusjr: Ideally it should, but I’m less sure when you consider providers can implement candidates incorrectly and induce all sorts of faulty states. I think we should rewrite the loop to better accomodate these sort of implementation errors and emit a better error message. Right now if something wrong happens, this would fail with `IndexError` after all states are popped, which is not very friendly for debugging.

src/resolvelib/resolvers.py
```diff
@@ -283,22 +283,39 @@ def _backtrack(self):
...
+                # Retrieve the last candidate pin and known incompatibilities.
+                broken_state = self._states.pop()
+                name, candidate = broken_state.mapping.popitem()
+                current_dependencies = {
+                    self._p.identify(d) for d in self._p.get_dependencies(candidate)
+                }
+                incompatible_deps = {
+                    self._p.identify(c.parent) for c in causes
+                }  # Parent should never be null
```
frostming, line 314: May need to move it out of the loop since it doesn't change

src/resolvelib/resolvers.py
```diff
@@ -283,22 +283,39 @@ def _backtrack(self):
...
-            name, candidate = broken_state.mapping.popitem()
+            # Ensure to backtrack to a state that caused the incompatibility
+            incompatible_state = False
+            while not incompatible_state:
+                # Retrieve the last candidate pin and known incompatibilities.
+                broken_state = self._states.pop()
+                name, candidate = broken_state.mapping.popitem()
+                current_dependencies = {
+                    self._p.identify(d) for d in self._p.get_dependencies(candidate)
+                }
```
frostming, line 311: I am afraid this is a waste to call `get_dependencies()` since it may be time-consuming. Maybe better to carry the info of (affected criteria names) of this round in the state, thoughts? @uranusjr @pradyunsg

frostming (approved):

src/resolvelib/resolvers.py
bennati: I improved the error message, is this what you had in mind?

src/resolvelib/resolvers.py
```diff
@@ -302,6 +302,9 @@ def _backjump(self, causes):
 
             # Ensure to backtrack to a state that caused the incompatibility
             incompatible_state = False
+            incompatible_deps = {
+                self._p.identify(c.parent) for c in causes
```
frostming, line 306: I am a still confused why it is `c.parent` not `c.requirement`. Let me show in an example, in the example, capitals are *candidates* while lowercases are *requirements*

```
State 1: A1 -> b1
State 2: C1 -> d1
State 3: E1 -> b2 and b1 conflicts with b2
```
Where `causes = [(b1, A1), (b2, E1)]`
We want to backjump to State 1, given:
```
get_dependencies(A1) = [b1]
```
this can be done by comparing the causes with the dependencies, shouldn't we use the .requirement attribute?

bennati: The actual example I've been working with has 3 requirements:

1. boto3
2. s3fs
3. seaborn

So the states are the following:
```
State 1: NONE -> boto3
State 2: NONE -> s3fs
State 3: NONE -> seaborn
State 4: BOTO3 -> botocore
...
State 7: S3FS -> aiobotocore
...
State 11: SEABORN -> matplotlib
```

And `causes = [(botocore, BOTO3), (botocore, aiobotocore)]`

So, the incompatibility is caused by a dependency of one requirement and a transitive dependency of another requirement.

The function `get-dependencies` returns only the direct dependencies, so basing this on requirements only would cause an infinite loop.
Does it make sense?

frostming: IIUC, `causes = [(botocore, BOTO3), (botocore, aiobotocore)]`, so we want to backjump to STATE 4 and mark BOTO3 as incompatible. To achieve this, we start from the **requirement name**(1) of causes, which is `botocore`, and traverse back the state stack, for each state, we call `get_dependencies()` on the new pin and see if the dependencies overlap with (1). So `get_dependencies(BOTO3) = [botocore]`, then we stop at STATE 4 and pop it from the stack.

If using `c.parent`, then (1) becomes [BOTO3, aiobotocore], how can we find the correct state?

frostming: Oh, you want to backjump to State 7, then it makes sense

frostming: Please also fix the lint errors to see if the CI is fine. And I will leave it to the other reviewers to decide whether this fix is valid. Thanks.

bennati: The state `AIOBOTOCORE -> botocore` does not exist, because a state without a pin is not added to the states (see [here](https://github.com/sarugaku/resolvelib/blob/7b66e2da3c1d0189d0ccd85b68bb761d843754b1/src/resolvelib/resolvers.py#L428))

I tested with `c.requirement` and the code crashes.

One solution would be to append the state, even if unpinned, and remove [this line](https://github.com/sarugaku/resolvelib/blob/7b66e2da3c1d0189d0ccd85b68bb761d843754b1/src/resolvelib/resolvers.py#L297), any thoughts?

uranusjr:
You can use `TYPE_CHECKING` to conditionally import those, so they are only imported for type checking.

bennati:
@uranusjr I used `TYPE_CHECKING` but that did not fix the issues in the tests, can you please review my last commit and let me know if something is wrong?

bennati:
Hi @uranusjr @frostming, I need to have this fixed merged soon, can you please let me know what else is required from my end? thanks!

frostming:
@bennati I would like to see the results of the CI, but it is blocked by #114 . I will get to it in the next week.

bennati:
@frostming I saw that #114 got merged, but we still have the issues with "unused" imports. 
Using `TYPE_CHECKING` did not help, unless there is something wrong in my code.

frostming:
@bennati you didn't merge the master branch, did you?

frostming:
Hmm, the result is as expected. This indicates that this patch does reduce resolution time in some cases, but also causes resolution failures since valid states are skipped.

bennati:
I would like to add tests for the following requirements
```
boto3==1.10.16
s3fs<0.5
seaborn
```

and
```
boto3==1.10.16
s3fs<0.5
botocore
```

I tried adding those in `tests/functional/python/inputs/case` but I do not understand how to generate a corresponding index, any help is appreciated!

frostming (approved):
LGTM now. It considers both requirements and parents.

abersheeran:
Anxiously waiting for the merge! This seems to solve the problem I'm having now! 😭

notatallshaw:
Hi all, I did some testing with this patch, very excited. I did have to comment out `self._r.rejecting_candidate(e.criterion, candidate)`, which I believe is a new method requirement on the reporter class and therefore should be highlighted in the release notes.

First some exciting news:
 * https://github.com/pypa/pip/issues/11760 with Python 3.7 resolves quickly
 * https://github.com/pypa/pip/issues/11480 with Python 3.9 resolves quickly

However when I ran Python 3.8 with `pip download apache-airflow[all]==1.10.13` I got the following error:

```
ERROR: Exception:
Traceback (most recent call last):
  File "/home/damian/backtrackfreeze/pip/src/pip/_internal/cli/base_command.py", line 160, in exc_logging_wrapper
    status = run_func(*args)
  File "/home/damian/backtrackfreeze/pip/src/pip/_internal/cli/req_command.py", line 247, in wrapper
    return func(self, options, args)
  File "/home/damian/backtrackfreeze/pip/src/pip/_internal/commands/download.py", line 138, in run
    requirement_set = resolver.resolve(reqs, check_supported_wheels=True)
  File "/home/damian/backtrackfreeze/pip/src/pip/_internal/resolution/resolvelib/resolver.py", line 92, in resolve
    result = self._result = resolver.resolve(
  File "/home/damian/backtrackfreeze/pip/src/pip/_vendor/resolvelib/resolvers.py", line 526, in resolve
    state = resolution.resolve(requirements, max_rounds=max_rounds)
  File "/home/damian/backtrackfreeze/pip/src/pip/_vendor/resolvelib/resolvers.py", line 410, in resolve
    name = min(unsatisfied_names, key=self._get_preference)
  File "/home/damian/backtrackfreeze/pip/src/pip/_vendor/resolvelib/resolvers.py", line 194, in _get_preference
    return self._p.get_preference(
  File "/home/damian/backtrackfreeze/pip/src/pip/_internal/resolution/resolvelib/provider.py", line 134, in get_preference
    candidate, ireqs = zip(*lookups)
ValueError: not enough values to unpack (expected 2, got 0)
```

Is the provider class expected to handle this new possibility or is this an error in the PR?

pradyunsg:
Doing running pip with `--debug` provide more context on how/why this error occured?

notatallshaw:
> Does running pip with `--debug` provide more context on how/why this error occured?

Here you go: https://gist.github.com/notatallshaw/ab6f6bee5f1f667838efed5168d3d125

notatallshaw:
I've never fully grokked resolvelib so excuse me if I'm misunderstanding, but the effect I see walking through this example is that at backtrack preference time there are unsatisfied requirements that do not have any install candidates.

On the assumption that this is a side effect of the backjumping, and this indicates that this should not be a preference, I changed the `get_preference` method of the provider class to return a tuple of infinities when `zip(*lookups)` throws a ValueError.

Perhaps a more elegant solution is if resolvelib can identify only unsatisfied requirements that have a candidate install requirement? Otherwise this definitely needs to be documented by the release notes and how exactly to handle it by the downstream provider class.

With my patch to get_preference  `apache-airflow[all]==1.10.13)` resolved in a fairly quick amount of time 🥳.

I will continue testing known real world problematic requirements.

notatallshaw:
For those who want to reproduce this error I identified but find the requirement `apache-airflow[all]==1.10.13` to be too difficult to run I found running this set of requirements (at least on Python 3.9 on Linux) to also produce the same error:

```
darglint
flake8
flake8-bandit
flake8-black
flake8-bugbear
flake8-builtins
flake8-comprehensions
flake8-docstrings
flake8-eradicate
flake8-isort
flake8-pytest-style
```

Taken from this real world example: https://github.com/pypa/pip/issues/11212#issuecomment-1332796541. (As a side note, once this lands I wonder if it's worth going back to a few of these old threads and someone who has the power to post in a locked thread let users know they should try the newest version of Pip).

frostming:
> Taken from real world example: [pypa/pip#10884 (comment)](https://github.com/pypa/pip/issues/10884#issue-1125051775)

I tried this minimal example and it resolves fast. It depends on the implementation of `get_preferences`, in the recent versions of `pip` BFS is used so that won't cause a problem

notatallshaw:
> > Taken from real world example: [pypa/pip#10884 (comment)](https://github.com/pypa/pip/issues/10884#issue-1125051775)
> 
> I tried this minimal example and it resolves fast. It depends on the implementation of `get_preferences`, in the recent versions of `pip` BFS is used so that won't cause a problem

I think you are missing the point of that issue, it isn't saying that the requirement takes a objectively long time to install it's saying that the requirement:

```
flake8
hacking
```

Takes **longer** and installs older versions than the requirement

```
hacking
flake8
```

Which is still true as of pip 23.0, at least as of my testing today.

But all of that is irrelevant to this PR, what is relevant is this PR causes Pip to throw an exception in `get_preferences` with the first set of requirements.

pradyunsg:
It’s fine if pips current implementation of get_preferences can’t handle backjumping — what’s important is that we should be able to change pip when we pull in changes to resolve lib into pip.

src/resolvelib/resolvers.py
```diff
@@ -283,22 +283,44 @@ def _backtrack(self):
...
-            name, candidate = broken_state.mapping.popitem()
+            # Ensure to backtrack to a state that caused the incompatibility
+            incompatible_state = False
+            while not incompatible_state:
+                # Retrieve the last candidate pin and known incompatibilities.
+                try:
+                    broken_state = self._states.pop()
+                    name, candidate = broken_state.mapping.popitem()
+                except (IndexError, KeyError):
+                    raise ResolutionImpossible(causes)
```
pradyunsg, line 315: Shouldn’t this instead be a case of raising an implementation error or RuntimeError or something?

pradyunsg: Meh, raising ResolutionImpossible seems right since we’d hit this if we aren’t able to backjump to a relevant state.

pradyunsg (approved):

pradyunsg:
This is quite nice @bennati! Thanks for poking at it and being willing to take feedback and act on it!

bennati:
Thank you @pradyunsg ! Happy to having been able to contribute. Is anything else required on my end before we can merge this?

pradyunsg:
A `news/{pr-number}.feature.rst` containing a changelog entry that’s user-facing (i.e. doesn’t discuss the internal concepts like state and whatnot) mentioning backjumping and that this will help make the resolver more efficient would be nice!

If you don’t get to that, that’s fine. One of maintainers can add that separately. :)

pradyunsg:
I've gone ahead and squashed the changes and added more context to the commit message; as well as adding the news fragment. Assuming the CI is happy, I'll land this. :)

pombredanne:
@pradyunsg @frostming Thanks!
@bennati you are awesome.

notatallshaw:
> Because what would it mean if the downstream library selected such a name as the most preferred?

It occured to me this is fairly easy to create a PoC to see if this actually causes resolvelib to fail and not just the downstream provider. I'll implement this evening NYC time after work.

pradyunsg:
Actually, @notatallshaw the issue you're reporting here seems to have been introduced in 0.9.0 -- I've just reproduced that in pip + resolvelib 0.9.0 in addition to this branch. My guess is that https://github.com/sarugaku/resolvelib/pull/111 is where we introduced the regression (which is overzealous trimming of relevant `information`).

pradyunsg:
> If this is not too much to ask making a release candidate would help a lot testing downstream.

Yes, and, you can do `pip install https://github.com/sarugaku/resolvelib/archive/refs/heads/main.zip` if you really want the latest and greatest. :)

notatallshaw:
> Actually, @notatallshaw the issue you're reporting here seems to have been introduced in 0.9.0 -- I've just reproduced that in pip + resolvelib 0.9.0 in addition to this branch. My understanding is that #111 is where we introduced the change (which is trimming `information` aggresively).

Agreed, I've opened a new issue with a reproducible set of steps: https://github.com/sarugaku/resolvelib/issues/120

frostming:
Hi, @pradyunsg @uranusjr do you have any preference on when a new release will be out with this fix?

pradyunsg:
Go ahead and cut the release. :)

uranusjr:
Still a few hours to 24 but I’ll call it.

notatallshaw:
FYI, I've raised a PR to remove backjumping: https://github.com/sarugaku/resolvelib/pull/142

PR #115, opened 2023-01-13
CI: fix the CI failures on packaging > 22

### LINKED ISSUE #114: Test Failures With Packaging 22.0/23.0
Resolvelib has multiple test failures when the tests are run using packaging 22.0 or 23.0.  All tests pass with 21.3.  It appears that packaging has implemented a more strict definition of what an appropriate version number is.  Here's the test summary:

```
======================================================== short test summary info ========================================================
FAILED tests/functional/cocoapods/test_resolvers_cocoapods.py::test_resolver[shared_parent_dependency] - packaging.version.InvalidVersion: Invalid version: '0.10.0-qs.0'
FAILED tests/functional/cocoapods/test_resolvers_cocoapods.py::test_resolver[deep_complex_conflict] - packaging.version.InvalidVersion: Invalid version: '4.0.0-preview2.1'
FAILED tests/functional/cocoapods/test_resolvers_cocoapods.py::test_resolver[shared_parent_dependency_with_swapping] - packaging.version.InvalidVersion: Invalid version: '0.3.6-pre.f7e3358'
FAILED tests/functional/cocoapods/test_resolvers_cocoapods.py::test_resolver[spapping_and_rewinding] - packaging.version.InvalidVersion: Invalid version: '2.0.0.pre.develop.2'
FAILED tests/functional/cocoapods/test_resolvers_cocoapods.py::test_resolver[pruned_unresolved_orphan] - packaging.version.InvalidVersion: Invalid version: '0.18.0-gh.de28323'
FAILED tests/functional/cocoapods/test_resolvers_cocoapods.py::test_resolver[conflict_common_parent] - packaging.version.InvalidVersion: Invalid version: '5.0.0-beta1.1'
================================================ 6 failed, 40 passed, 3 xfailed in 6.87s
```

See the attached for the details.

Scott K
[test_results_packaging_22_0.txt](https://github.com/sarugaku/resolvelib/files/10401881/test_results_packaging_22_0.txt)

frostming:
Because there are test cases against indexes for other languages and their version don't follow the python standards, ~~we have to set version cap of packaging to `<22.0` for testing.~~

Also, switch to `nox` in CI workflows.

* * *
Fixes #114

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -44,7 +44,7 @@ lint =
 	types-requests
 test =
 	commentjson
-	packaging<22.0  # to handle non-python standard versions
+	packaging
 	pytest
 release =
 	build
--- a/setup.cfg
+++ b/setup.cfg
@@ -57,7 +57,7 @@ universal = 1
 [flake8]
 max-line-length = 88
 select = C,E,F,W,B
-ignore = E203, W503
+ignore = E203, W503, F401
 exclude =
 	.git,
 	.venv,
--- a/setup.cfg
+++ b/setup.cfg
@@ -44,7 +44,7 @@ lint =
 	types-requests
 test =
 	commentjson
-	packaging
+	packaging<22.0  # to handle non-python standard versions
 	pytest
 release =
 	build
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
       - run: nox -s release -- --version '' --repo '' --prebump ''
   test:
     name: Test
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     needs: [lint]
     strategy:
       fail-fast: true
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,6 @@ jobs:
       fail-fast: true
       matrix:
         python:
-          - "2.7"
           - "3.10"
           - "3.9"
           - "3.8"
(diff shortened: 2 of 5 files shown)


frostming:
Python 2.7 is failing due to incompatible `nox`. Since we are going to drop 2.7(#116) I prefer to remove it from the matrix.

kitterma:
What's the solution for those that don't have the option to stay with the old version?

frostming:
> What's the solution for those that don't have the option to stay with the old version?

That is the test dependency and resolvelib itself doesn't have any dependency

kitterma:
Right, but the environment I'm looking at (Debian Testing) has the newer version, so running the tests with the older version isn't an option.  Maybe put the relevant tests inside try/except and then pass on that error from packaging so testing with the newer version at least doesn't cause inappropriate test failures?

Scott K

uranusjr:
The Swift tests actually implement a hand-rolled extremely simplified parser, perhaps we could do the same? https://github.com/sarugaku/resolvelib/blob/master/tests/functional/swift-package-manager/test_resolvers_swift.py

setup.cfg
```diff
@@ -57,7 +57,7 @@ universal = 1
 [flake8]
 max-line-length = 88
 select = C,E,F,W,B
-ignore = E203, W503
+ignore = E203, W503, F401
```
uranusjr, line 60: Where is F401 triggered?

frostming: `typing` members that only appear in comments.

uranusjr: Ah OK.

PR #126, opened 2023-03-08
fix: use the protocol method instead of .name attribute

frostming:


pradyunsg (approved):

pradyunsg:
@uranusjr wanna cut 1.0.1 for this?

Issue #116, opened 2023-01-13
Drop support for Python prior to 3.7


frostming:
I think it's time to drop support for Python versions prior to 3.7.

/cc @uranusjr @pradyunsg

uranusjr:
Yes

uranusjr:
This is mostly plain logic and the 2/3 difference is close to none. Perhaps the only thing that matters in typing?

frostming:
> Perhaps the only thing that matters is typing?

Yes, and CI infrastructures don't support 2.7 well.

frostming:
Duplicate with #98

PR #129, opened 2023-03-17
Declare supported Python versions in dist meta

### LINKED ISSUE #122: Required Python versions not specified in pip-readable way
That causes every version of the package to be installed on every Python version.

Could you please add a `python_requires` property in your setup.cfg?

OlenaYefymenko:
Resolvelib is not maintained for Python prior to 3.7. It was pointed out in #116.

Resolves #122

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,7 +26,7 @@ package_dir =
 packages = find:
 include_package_data = true
 zip_safe = false
-python_requires = >=2.7, !=3.0, !=3.1, !=3.2, !=3.3, !=3.4, !=3.5, !=3.6
+python_requires = >=3.7
 
 [options.packages.find]
 where = src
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -26,6 +26,7 @@ package_dir =
 packages = find:
 include_package_data = true
 zip_safe = false
+python_requires = >=2.7, !=3.0, !=3.1, !=3.2, !=3.3, !=3.4, !=3.5, !=3.6
```
pradyunsg, line 29:
```suggestion
python_requires = >=3.7
```

We don't really need to worry about Python 2 support now. 8 years of extended support from volunteers is more than enough.

webknjaz: Readability nit: I'd add a whitespace between the comparison operator and the version number if you don't mind.

```suggestion
python_requires = >= 3.7
```

Issue #131, opened 2023-05-29
Option to only provide causes of conflict when calling get_preference


notatallshaw:
When https://github.com/sarugaku/resolvelib/pull/84 was implemented it gave the downstream library the information about whether something was the cause of a backtracking conflict or not. 

However using this information is problematic from a performance perspective as it can create an O(n<sup>2</sup>) situation as the entire list of names is checked over and for each name the entire list of backtrack causes is checked over. This has created real world performance issue https://github.com/pypa/pip/pull/10621

There have been of a couple of unmerged PRs to attempt to fix this via either by doing some fancy caching logic in Pip or defining a formal object structure for what a "Cause" object should look like and then add cache on to that object.

However a much simpler solution is to take the implicit loop out of `get_preference` and when there is a conflict to only pass in names that are part of the backtrack cause. It seems to be this information is sufficiently generic for a resolving library that it can be an option of the algorithm itself and not just a preference of the downstream library.

I will start to make a PR and see how it performs, I opened this issue now to see if there is any direct objections to this approach.

notatallshaw:
Initial testing for this looks good, I consistently see a 2 second save on running `python -m pip download apache-airflow[all]==2.6.1` (with all downloads cached) or slightly about 1% of the total time, which might not sound like much but most of the time is being spent building packages in isolated environments to extract the metadata so the fact the performance improvement is measurable and consistent is a big win here.

But actually I think there's an even bigger improvement from this approach, it allows to set a reporting event that lets the downstream library log the cause of backtracking, I think this will be very useful for users trying to debug why backtracking is happening in the first place.

I'll try and post a PR soon.

notatallshaw:
My initial approach, https://github.com/sarugaku/resolvelib/pull/132, made incorrect assumptions about what information resolvelib could guarantee it has access to.

This would now be solved by https://github.com/sarugaku/resolvelib/pull/132, the provider would need to implement on their side which causes to filter out.

Issue #64, opened 2020-12-08
“Learn” to avoid trying routes that do not affect the dependency graph


uranusjr:
```
pip install "apache-airflow[devel_ci]==1.10.14rc1" --constraint "https://raw.githubusercontent.com/apache/airflow/constraints-1-10/constraints-3.6.txt"
```

This (run against https://github.com/pypa/pip/commit/fd8ddb6e55086500b1d836b07365ba7404f53e3b) would result in the resolver backtracking on setuptools, but these backtracks are obviously useless since setuptools does not have dependencies.

The resolver should be able to “look ahead” and see that trying other versions of setuptools would not change the conflicting graph, and skip working on them altogether. This applies not only to packages without dependencies (although they are the easiest target), but also versions that have the same dependencies. The idea of “the same dependencies” however require provider support (a new hook `are_dependencies_equal` or something) so those are more difficult to implement.

pfmoore:
I'm not sure I understand. You say "setuptools does not have dependencies", but surely we can only say that "this *specific* version of setuptools does not have dependencies"? If we backtrack and try a different version of setuptools, we might get one that does have dependencies.

Am I missing something? (It's early and I haven't had a cup of tea yet, so it's quite likely 🙂)

uranusjr:
Yes, sorry for the lax language, I was in a hurry writing the message. What I have in mind is the resolver should be able to look at (say) setuptools 40, and realise the rest of the dependency graph after pinning it would be exactly the same as setuptools 40.1. So if 40.1 failed, 40.0 is also destined to fail, and should be skipped entirely.

pfmoore:
Ah, OK. So if we fail on 40.1, then the resolver can ask the provider "does 40.0 have the same dependencies?" and ignore 40.0 if the answer is "no".

I can see how that would help, but would it not be almost as good (and not require an additional mechanism) if the provider cached dependency data in memory, so that checking 40.0 would be fast enough that skipping it is not worth it? That would mean breaking pathological cases like projects that calculate dependencies randomly, but IMO it's fine for a provider to make that choice (and I'd be fine with pip doing so, as we already cache wheels which in effect does that).

Of course, it's not an either/or choice - we can do both.

pradyunsg:
This is essentially the CDCL style behaviour - I'm not a 100% sure I understand how we'd phrase that in terms of criterion-based model we have here, but I'm a strong +1 on this.

uranusjr:
[...]

PR #132, opened 2023-06-03
Always prefer causes when backtracking

notatallshaw:
As per https://github.com/sarugaku/resolvelib/issues/131 this moves the responsibility from the downstream library to prefer backtracking causes into resolvelib itself solving the issue of an O(n<sup>2</sup>) performance problem that can happen.

It provides additional performance benefits under those circumstances because `get_preference` will be called a lot less.

After working on this PR I decided not to have this as optional, it keeps the resolver and tests much simpler and while it may be possible to construct a dependency graph where this approach is slower we have no such reports come from Pip and as per [this comment](https://github.com/sarugaku/resolvelib/issues/64#issuecomment-1567302535) in real world benchmarking there is over a 2x improvement in resolution speed to preferring backtracking, this makes intuitive sense as any successful solution must not have conflicts so resolving them first will generally be faster.

But let me know if you would prefer this approach to be optional, I can do so.

I have been testing these change in Pip on this branch: https://github.com/pypa/pip/compare/main...notatallshaw:pip:always-prefer-causes


src/resolvelib/resolvers.py
```diff
@@ -418,10 +417,29 @@ def resolve(self, requirements, max_rounds):
...
+                )
+                if unsatisfied_causes_names:
+                    unsatisfied_names = list(unsatisfied_causes_names)
+
+                self._r.backtracking_on(
+                    backtrack_cause_names, unsatisfied_causes_names
+                )
+
             # Choose the most preferred unpinned criterion to try.
             name = min(unsatisfied_names, key=self._get_preference)
```
notatallshaw, line 444: I could put a check here to see if the length of unsatisfied names is 1 and skip calling `self._get_preference` altogether under this scenario?

src/resolvelib/providers.py
```diff
@@ -9,14 +9,7 @@ def identify(self, requirement_or_candidate):
...
 
-    def get_preference(
-        self,
-        identifier,
-        resolutions,
-        candidates,
-        information,
-        backtrack_causes,
-    ):
+    def get_preference(self, identifier, resolutions, candidates, information):
```
uranusjr, line 12: We’re at version 1 now so it’d be better if we could implement this in a backward-compatible way. Maybe we can still pass the backtrack causes anyway?

uranusjr:
Logic looks good, one question about interface change. Would it be possible to have a test for this?

src/resolvelib/providers.py
notatallshaw: Sure, it was just to simplify the interface and make sure downstream libraries aren't duplicating this logic, but I don't have any strong opinions on whether to keep providing it or not.

notatallshaw:
> Logic looks good, one question about interface change. Would it be possible to have a test for this?

I will look at the existing tests and see if I can sufficiently adapt one.

notatallshaw:
I'm closing this, this optimization is relatively niche in real world (only one reported case of significant slowdown and never been able to reproduce).

But I do have a bigger optimization that will extend this strategey to "always prefer conflicting causes, then causes", I beleive that optimization, so this approach will be included in that and have stronger justification with lots of real world examples I can show it speeds up. I will make a new PR for that one.

PR #136, opened 2023-08-30
add required arg `backtrack_causes` to `get_preferences()` override

jimkring:
backtrack_causes is a required argument in get_preferences, as defined by the parent class resolvelib.AbstractProvider. Adding this argument allows the example to run without errors.

### DIFF
--- a/examples/reporter_demo.py
+++ b/examples/reporter_demo.py
@@ -73,7 +73,7 @@ class Provider(resolvelib.AbstractProvider):
     def identify(self, requirement_or_candidate):
         return requirement_or_candidate.name
 
-    def get_preference(self, identifier, resolutions, candidates, information):
+    def get_preference(self, identifier, resolutions, candidates, information, backtrack_causes):
         return sum(1 for _ in candidates[identifier])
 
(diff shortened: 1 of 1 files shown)


jimkring:
not sure why CI / Lint failed.  nox/black check works for me :)

frostming:
> not sure why CI / Lint failed. nox/black check works for me :)

It complains because the line should be broken into one argument per line

examples/reporter_demo.py
```diff
@@ -73,7 +73,7 @@ def __init__(self, index):
     def identify(self, requirement_or_candidate):
         return requirement_or_candidate.name
 
-    def get_preference(self, identifier, resolutions, candidates, information):
+    def get_preference(self, identifier, resolutions, candidates, information, backtrack_causes):
```
frostming, line 76:
```suggestion
    def get_preference(
        self,
        identifier,
        resolutions,
        candidates,
        information,
        backtrack_causes,
    ):
```

uranusjr (approved):
Maybe we should add a CI job to ensure the demo runs

PR #138, opened 2023-09-08
Clarify the docstrings for `providers.py`

### LINKED ISSUE #137: Awkward wording in docstring for `find_matches()`
The docstring says the `identifier` parameter "identifies the dependency matches of which should be returned." Is this trying to say it "it identifies the dependency for which matches should be returned"?

brettcannon:
Closes #137


pradyunsg (approved):
LGTM, but I'll wait for @uranusjr since it's definitely possible that there's a gap in my understanding. :)

PR #140, opened 2023-10-04
Looking for feedback on narrow backtrack causes approach

notatallshaw:
This PR is an approach to https://github.com/sarugaku/resolvelib/issues/139 and fixes https://github.com/pypa/pip/issues/12305#issuecomment-1743976704

However, as the test suite shows it's using methods that are not guaranteed to exist. 

I am looking for feedback on how this could be implemented in a way that does not create brand new requirements on downstream libraries.

### DIFF
--- a/src/resolvelib/resolvers.py
+++ b/src/resolvelib/resolvers.py
@@ -267,6 +267,34 @@ class Resolution(object):
         # end, signal for backtracking.
         return causes
 
+    def _narrow_causes(self, causes):
+        """Return a narrowed causes list
+        """
+        # Causes are often duplicates, first dedup them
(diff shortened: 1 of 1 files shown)


notatallshaw:
Actually I realized, as I woke up this morning, the solution is quite obvious.

As only the downstream library can know the shape of their own requirements then this must be implemented there with resolvelib simply making a call to ask the library to narrow the causes if it can.

Once I get a chance I will make an issue on Pip side, and once agreed this is a good idea I will edit this PR to simply call a narrow causes method on the downstream library.

notatallshaw:
I've created an issue on Pip side (https://github.com/pypa/pip/issues/12318), once I have feedback there I will make a clean PR here.

PR #144, opened 2023-11-29
Add intelligent fallback from backjump to backtrack

notatallshaw:

Alternative to https://github.com/sarugaku/resolvelib/pull/142, fixes both the requirments mentioned here https://github.com/sarugaku/resolvelib/pull/142#issuecomment-1811627400

I beleive this strategy is fairly foolproof:

1. When backjumping make a backup of the states the first time backjumping skips over states further than backtracking would
2. If backjumping reaches resolution impossible and a backup state exists, fall back to that state and start backtracking

However, there are downsides, it could definetly cause a valid ResolutionImpossible to take much longer to produce. I have therefore give the calling library a choice on what to do, backjump with backtrack fallback, backjump only, or backtrack only.

If you want tests I need some small guidance, I don't really understand this repos tests.

Let me know what you think!

### DIFF
--- a/src/resolvelib/resolvers.py
+++ b/src/resolvelib/resolvers.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import collections
 import itertools
 import operator
+from copy import copy
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -122,11 +123,7 @@ class Resolution(Generic[RT, CT, KT]):
         coming round.
         """
         base = self._states[-1]
-        state = State(
-            mapping=base.mapping.copy(),
-            criteria=base.criteria.copy(),
-            backtrack_causes=base.backtrack_causes[:],
-        )
+        state = copy(base)
         self._states.append(state)
 
     def _add_to_criteria(
@@ -274,14 +271,7 @@ class Resolution(Generic[RT, CT, KT]):
 
     def _copy_all_states(self) -> list[State]:
         """Create a copy of the all the current states"""
-        return [
-            State(
-                s.mapping.copy(),
-                s.criteria.copy(),
-                s.backtrack_causes[:],
-            )
-            for s in self._states
-        ]
+        return [copy(s) for s in self._states]
 
     def _backtrack_iteration(self) -> tuple[KT, CT, list[tuple[KT, list[CT]]]]:
         """
--- a/src/resolvelib/resolvers.py
+++ b/src/resolvelib/resolvers.py
@@ -11,6 +11,7 @@ from typing import (
     Iterable,
     Mapping,
     NamedTuple,
+    Optional,
 )
 
 from .providers import AbstractProvider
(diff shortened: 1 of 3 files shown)


src/resolvelib/resolvers.py
```diff
@@ -9,8 +9,12 @@
     Collection,
     Generic,
     Iterable,
+    List,
```
frostming, line 12: You can use `list` and `tuple` directly since `annotations` has been imported.

notatallshaw: Fixed

notatallshaw:
Two thoughts now I've had time to sleep on this.

Firstly, adding the option for the calling library to choose a behavior is perhaps a bit too much. I'm thinking perhaps remove it and see if calling libraries notice an issue?

Secondly this fallback approach could be extended further. It currently does: Backjumping -> Resolution Impossible -> Backtrack forever. It could do Backjumping -> Resolution Impossible -> Backtrack Once -> Backjumping. But I think it's better as it is now and if this lands and if we see real world instances where fallback causes significant slow downs it could be considered.

notatallshaw:
> Firstly, adding the option for the calling library to choose a behavior is perhaps a bit too much. I'm thinking perhaps remove it and see if calling libraries notice an issue?

I've created a seperate commit which applies this so the two versions can be compred. Depending on preference I can delete or squash the commit.

notatallshaw:
FYI a seperate Pip PR (https://github.com/pypa/pip/pull/12459) I have removes all examples I have tested for backjumping incorrectly producing `ResolutionImpossible`, I think it's important to still remove or have a fallback for backjumping, I have a more detailed comment in the alternative PR https://github.com/sarugaku/resolvelib/pull/142#issuecomment-1880145310

frostming (commented):
The implementation is a bit convoluted, need more comments to help successors understand what is happening.

notatallshaw:
> The implementation is a bit convoluted, need more comments to help successors understand what is happening.

I think to a certain extent that is the nature of this solution, it backs up states and only uses those backed up states under the right conditions.

But I have tried to improve the situation breaking up the code a little, adding more comments, and trying to pick better names. Let me know what you think.

notatallshaw:
An further thoughts on this?

It would be good to fix this backjump issue somehow, as it breaks real world requirements for both pip and PDM.

src/resolvelib/resolvers.py
```diff
@@ -269,6 +272,89 @@ def _attempt_to_pin_criterion(self, name: KT) -> list[Criterion[RT, CT]]:
...
         return causes
 
+    def _copy_all_states(self) -> list[State]:
+        """Create a copy of the all the current states"""
+        return [
+            State(
+                s.mapping.copy(),
+                s.criteria.copy(),
+                s.backtrack_causes[:],
+            )
```
uranusjr, line 282: Maybe we should just implement the copy protocol on State.

notatallshaw: Done

notatallshaw:
I am unhappy with this solution, I am working further on understanding the root cause of the problem: https://github.com/sarugaku/resolvelib/issues/134#issuecomment-2164347177

I may recreate this PR once I have a working test.

PR #145, opened 2023-12-01
Allow provider to filter unsatisfied names, when backtracking

notatallshaw:
This allows the provider to optimize the steps it takes to choose what to backtrack on.

Specifically the aim is to add preferring conflicting causes in Pip without the overhead of recalculating or caching each time `get_preference` is called. I initially wanted to do a lot of this "preferring" work in resolvelib itself, but upon inspecting the resolution types I realized there was no guarantee of required information (which would have been a problem if this previous PR had landed https://github.com/sarugaku/resolvelib/pull/132).

I have a Pip branch that implements this filtering: https://github.com/notatallshaw/pip/blob/prefer-conflicting-causes/src/pip/_internal/resolution/resolvelib/provider.py#L240. Which on Python 3.8 allows it to solve the following requirement which currently produces a "ResolutionTooDeep" error: `--only-binary ":all:" "numpy==1.21.6" "cython==0.29.28" "scipy>=1.4.0" "torch>=1.7" "torchaudio" "soundfile" "librosa==0.10.0.*" "numba==0.55.1" "inflect==5.6.0" "tqdm" "anyascii" "pyyaml" "fsspec>=2021.04.0" "aiohttp" "packaging" "flask" "pysbd" "pandas" "matplotlib" "trainer==0.0.20" "coqpit>=0.0.16" "pypinyin" "mecab-python3==1.0.5" "jamo" "bangla==0.0.2" "k_diffusion" "einops" "transformers"`

### DIFF
--- a/src/resolvelib/providers.py
+++ b/src/resolvelib/providers.py
@@ -40,6 +40,14 @@ class AbstractProvider(Generic[RT, CT, KT]):
     ) -> Preference:
         """Produce a sort key for given requirement based on preference.
 
+        As this is a sort key it will be called O(n) times per backtrack step,
+        where n is the number of `identifier`s. If you have a check which is
+        expensive in some sense, e.g. it needs to make O(n) checks per
+        identifier, or takes significant wall clock time but could be short
+        circuited once finding an identifier that matches the check, consider
+        using `narrow_requirement_selection` to filter the `identifier`s
+        before this sort key is called.
+
         The preference is defined as "I think this requirement should be
         resolved first". The lower the return value is, the more preferred
         this group of arguments is.
@@ -146,11 +154,8 @@ class AbstractProvider(Generic[RT, CT, KT]):
     ) -> Iterable[KT]:
         """
         An optional method to narrow the selection of requirements being
-        considered during resolution.
-
-        The requirement selection is defined as "The possible requirements
-        that will be resolved next." If a requirement is not part of the returned
-        iterable, it will not be considered during the next step of resolution.
+        considered during resolution. This method is called O(1) time per
+        backtrack step.
 
         :param identifiers: An iterable of `identifiers` as returned by
             ``identify()``. These identify all requirements currently being
@@ -174,19 +179,17 @@ class AbstractProvider(Generic[RT, CT, KT]):
           the requirement, or ``None`` to indicate a root requirement.
 
         Must return a non-empty subset of `identifiers`, with the default
-        implementation being to return `identifiers` unchanged.
+        implementation being to return `identifiers` unchanged. Those `identifiers`
+        will then be passed to the sort key `get_preference` to pick the most
+        prefered requirement to attempt to pin, unless `narrow_requirement_selection`
+        returns only 1 requirement, in which case that will be used without
+        calling the sort key `get_preference`.
 
-        Can be used by the provider to optimize the dependency resolution
-        process. `get_preference` will only be called for the identifiers
-        returned. If there is only one identifier returned, then `get_preference`
-        won't be called at all.
-
-        Serving a similar purpose as `get_preference`, this method allows the
-        provider to guide resolvelib through the resolution process. It should
-        be used instead of `get_preference` for logic when the provider needs
-        to consider multiple identifiers simultaneously, or when the provider
-        wants to skip checking all identifiers, e.g., because the checks are
-        prohibitively expensive.
+        This method is designed to be used by the provider to optimize the
+        dependency resolution, e.g. if a check cost is O(m) and it can be done
+        against all identifiers at once then filtering the requirement selection
+        here will cost O(m) but making it part of the sort key in `get_preference`
+        will cost O(m*n), where n is the number of `identifiers`.
 
         Returns:
             Iterable[KT]: A non-empty subset of `identifiers`.
--- a/src/resolvelib/providers.py
+++ b/src/resolvelib/providers.py
@@ -40,13 +40,12 @@ class AbstractProvider(Generic[RT, CT, KT]):
     ) -> Preference:
         """Produce a sort key for given requirement based on preference.
 
-        As this is a sort key it will be called O(n) times per backtrack step,
-        where n is the number of `identifier`s. If you have a check which is
-        expensive in some sense, e.g. it needs to make O(n) checks per
-        identifier, or takes significant wall clock time but could be short
-        circuited once finding an identifier that matches the check, consider
-        using `narrow_requirement_selection` to filter the `identifier`s
-        before this sort key is called.
+        As this is a sort key it will be called O(n) times per backtrack
+        step, where n is the number of `identifier`s, if you have a check
+        which is expensive in some sense. E.g. It needs to make O(n) checks
+        per call or takes significant wall clock time, consider using
+        `narrow_requirement_selection` to filter the `identifier`s, which
+        is applied before this sort key is called.
 
         The preference is defined as "I think this requirement should be
         resolved first". The lower the return value is, the more preferred
--- a/src/resolvelib/providers.py
+++ b/src/resolvelib/providers.py
@@ -135,3 +135,60 @@ class AbstractProvider(Generic[RT, CT, KT]):
         specifies as its dependencies.
         """
         raise NotImplementedError
+
+    def narrow_requirement_selection(
+        self,
+        identifiers: Iterable[KT],
+        resolutions: Mapping[KT, CT],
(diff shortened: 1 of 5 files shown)


notatallshaw:
Interestingly this seems to turn `tests/test_resolvers.py::test_resolving_conflicts` into a flaky test.

I think the only behavioral change that should have happened is that `get_preference` is not always called.

I will need to investigate if that's the issue or if it's something else.

notatallshaw:
I have fixed the flaky test.

The test relied on the order of resolution and this was broken by giving get_preference a set instead of a list. I got rid of keeping this set, as my code no longer relies on that anyway.

notatallshaw:
I have started to test the improvements using pip resolver branchmark: https://github.com/pradyunsg/pip-resolver-benchmarks

The default benachmark of pyrax==1.9.8, moderetly improves from 4.75 +- 0.01 to 4.62 +- 0.04 on my machine.

This benchmark is pretty interesting because it is a complicated resolve but doesn't benefit at all from choosing conflicting causes, so the extra checks for choosing conflicting causes in my experimental Pip branch contribute a little extra time.

The small performance improvement comes from deduplicating and filtering the causes ahead of calling `get_preference`.

notatallshaw:
I'm not really a fan of the name `filter_unsatisfied_names`, but I can't think of a better name right now.

notatallshaw:
Here is another example that produces `ResolutionTooDeep` in the current version of Pip:

```
sphinx
sphinx-rtd-theme
sphinx-toolbox
myst-parser
sphinxcontrib-bibtex
nbsphinx
```

But can be solved by implementing this in resolvelib and adding logic on Pip side to prefer conflicting causes. 

Let me know what would be needed to review and accept this, though like I'm sure you I won't be available much until the new year, I will hopefully be pushing for it a bit harder early next year.

As resolvelib doesn't ever implement this method itself I wasn't sure what to add in terms of tests.

notatallshaw:
Working on the Pip implementation (https://github.com/pypa/pip/pull/12459), I realized it can be useful to have more than just backtrack_causes, so I have updated the `filter_unsatisfied_names` to basically mirror the `get_preference` arguments.

The Pip implementation now improves things to the extent that I can no longer reproduce the `ResolutionImpossible` failures that backjumping causes. I will write about that more in the other issues/PRs.

notatallshaw:
I've updated this PR based on the conversation here: https://github.com/pypa/pip/issues/12497

Main changes are:

 * `filter_unsatisfied_names` -> `narrow_backtrack_selection`, still would like a better name, but I think it addresses the concerns raised in that thread
 * Significantly updated the docstring to explain the purpose of the method and how it differs to `get_preference`

src/resolvelib/providers.py
```diff
@@ -136,3 +136,36 @@ def get_dependencies(self, candidate: CT) -> Iterable[RT]:
...
+
+        Serving a similar purpose as `get_preference`, this method allows
+        the provider to guide resolvelib through the backtracking process.
+        This method should be used instead of `get_preference` when the
+        provider needs to consider multiple unsatisfied dependency names
+        simultaneously, or when the provider wants to skip checking all
+        names, e.g., because the checks are prohibitively expensive.
+
+        Returns:
+            Iterable[KT]: A non-empty subset of `unsatisfied_names`.
```
pfmoore, line 169: Could the docstring be extended to describe the meaning of the various arguments? For example, "information" doesn't really describe what the provider can expect to find in that value... 😉

notatallshaw: Oh yeah, since this method now matches `get_preference` almost entirely (it didn't when I opened this PR) I've updated the docstring to far more closely match `get_preference`, including pulling in the parameter descriptions from there.

I also updated the method name tp `narrow_requirement_selection`, as it fits the docstring descriptions better now, and I like that `backtrack` isn't part of the name (in fact I would like to move backtrack from the parameter name as well, but that would be a breaking change for `get_preference`, I was the one that added that parameter name but I wasn't thinking about resolvelib in a wider context when I added it).

src/resolvelib/resolvers.py
```diff
@@ -435,8 +435,33 @@ def resolve(
...
+                filtered_unstatisfied_names = unsatisfied_names
+
             # Choose the most preferred unpinned criterion to try.
-            name = min(unsatisfied_names, key=self._get_preference)
+            if len(filtered_unstatisfied_names) > 1:
+                name = min(
+                    filtered_unstatisfied_names, key=self._get_preference
+                )
+            else:
+                name = filtered_unstatisfied_names[0]
```
frostming, line 463: why need this if-else? `min` should return immediately if the length is 1

src/resolvelib/providers.py
```diff
@@ -136,3 +136,60 @@ def get_dependencies(self, candidate: CT) -> Iterable[RT]:
...
+        provider to guide resolvelib through the resolution process. It should
+        be used instead of `get_preference` when the provider needs to consider
+        multiple identifiers simultaneously, or when the provider wants to skip
+        checking all identifiers, e.g., because the checks are prohibitively
+        expensive.
+
+        Returns:
+            Iterable[KT]: A non-empty subset of `identifiers`.
+        """
+        raise NotImplementedError
```
frostming, line 195: I am afraid this will cause breaking changes, can we have a default implementation to return `identifiers` as-is?

src/resolvelib/resolvers.py
notatallshaw: `min` still calls then key function if there is 1 element, the point is to avoid an unnecessary call to `get_preference`

src/resolvelib/providers.py
notatallshaw: The only reason I didn't do this is because it was called an Abstract class, if you're okay with a default implementation I'll update this shortly.

src/resolvelib/resolvers.py
pfmoore: If that's the logic, it's worth a comment. (Also, I'd argue that `min` shouldn't do that, but I'm sure there's some pathological case out there that would break if it changed 🙁)

src/resolvelib/providers.py
notatallshaw: Done

src/resolvelib/resolvers.py
notatallshaw: Updated

examples/extras_provider.py
```diff
@@ -52,3 +52,13 @@ def get_dependencies(self, candidate):
             req = self.get_base_requirement(candidate)
             deps.append(req)
         return deps
+
+    def narrow_requirement_selection(
```
uranusjr, line 56: If I understand correctly, implementing this function is optional since it just returns all identifiers by default, right? It’d be good to make this clear in docs, and remove all these stub implementations in examples.

notatallshaw:
It occurs to me that resolvelib could validate that resolvelib could validate that the return value of `self._p.narrow_requirement_selection` is a subset of `unsatisfied_names`, this would guarantee that this method does not affect the soundness of the resolution, but at the cost of doing a subset check every backtrack iteration.

I am inclined to leave it to the documentation of how the client must implement `narrow_requirement_selection`.

src/resolvelib/resolvers.py
```diff
@@ -435,8 +435,34 @@ def resolve(
...
+                narrowed_unstatisfied_names = unsatisfied_names
+
+            # If there is only 1 unsatisfied name skip calling self._get_preference
+            if len(narrowed_unstatisfied_names) > 1:
+                # Choose the most preferred unpinned criterion to try.
+                name = min(
+                    narrowed_unstatisfied_names, key=self._get_preference
+                )
+            else:
+                name = narrowed_unstatisfied_names[0]
```
uranusjr, line 464: It’s worthwhile to add an explicit check to ensure `narrowed_unstatisfied_names` is not empty. User errors happen. Or maybe it’s viable to allow returning an empty list and do something sensible (what?) in the situation?

notatallshaw: I think there are two valid options here:

1. Explicit check and throw specific error
2. Explicit check and use unsatisfied_names instead of narrowed_unstatisfied_names

I don't mind which, it's a philisophical question of if the algorithm should continue with the information it has, or exit early with an error. I'm going to do number 1 for now, let me know if I should change it.

notatallshaw:
Okay, I've rebased, applied suggestions, updated doc strings to be more readable (at least to me), added a news entry, and extended Python functional tests to check with and without this method, and the test is a realistic use of this method.

Feeling pretty good about this now, looking for further feedback (or approval 🤞).

Issue #147, opened 2023-12-02
For n packages there are O(n<sup>2</sup>) calls to  `_is_current_pin_satisfying` 


notatallshaw:
As the number of packages to resolve increases there appears to be an O(n<sup>2</sup>) calls to  `_is_current_pin_satisfying` and I think, at least in some circumstances, it's possible to optimize this away.

I am using steps to reproduce from https://github.com/pypa/pip/issues/12320 and the branch from this PR to avoid network calls in the call graph https://github.com/pypa/pip/pull/12327 and not have O(n<sup>2</sup>) calls to Pip collecting packages from local directory. In the future I plan to be able to show this performance issuue using https://github.com/pradyunsg/pip-resolver-benchmarks,

In this example there are ~1300 packages to install and `_is_current_pin_satisfying`  is called ~2.5 million times, I generated the call graph using cProfile and gprof2dot:

![output](https://github.com/sarugaku/resolvelib/assets/8070352/3dbf1b79-fea6-40b8-83f3-f72d66b1bcda)

I have not yet investigated how one might optimize this yeat.

This is motivated from a real world usage report: https://github.com/pypa/pip/issues/12314.

notatallshaw:
An idea I will investigate if no one else pipes in, but maybe it's possible to keep track of satified and unsatisfied names in the state, rather than needing to recalculate them each time.

notatallshaw:
FYI I have not made any meaningful progress on this and doesn't expect to any time soon. 

I tried avoiding calling `_is_current_pin_satisfying` in most cases of calculating `unsatisfied_names` , but I got lots of test failures that I beleive to be valid, it feels like it should be possible so possibly there was just an unidentified bug in my code. I will take a look at this again when I have a chance.

notatallshaw:
FYI, a workaround is for the provider to cache the calls, a PR has been raised on Pip side to do that: https://github.com/pypa/pip/pull/12453

PR #148, opened 2023-12-03
Speed up resolution by approximating unsatisfied names

notatallshaw:
Fixes https://github.com/sarugaku/resolvelib/issues/147

Creating as a draft PR because it currently causes a lot of test failures. I don't understand yet if that's because this approach is ultimately unsound or just that some of the assumptions about how resolvelib works are broken, but resolvelib still always finds a correct resolution. If anyone has time to assist in understanding these test failures or explaining what they mean that would help, but otherwise I will figure them out.

Using @pradyunsg's https://github.com/pradyunsg/pip-resolver-benchmarks I see a significant resolution time improve of ~5.5x in the pyrax_198 scenario.

### DIFF
--- a/src/resolvelib/resolvers.py
+++ b/src/resolvelib/resolvers.py
@@ -212,18 +212,6 @@ class Resolution(Generic[RT, CT, KT]):
             backtrack_causes=self.state.backtrack_causes,
         )
 
-    def _is_current_pin_satisfying(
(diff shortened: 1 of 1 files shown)


notatallshaw:
Closing this, as implemented this isn't right.

I think it might be possible to salvage the optimization approach, I'm make another PR once/if I have one that passes the tests.

PR #149, opened 2023-12-09
Separate `AbstractResolver` and `Resolver` into different modules

pradyunsg:
I started looking into porting an existing SAT solver, but this lack of separation makes it non-trivial to organise the relevant code.

Making this change to simplify that, as well as create a more explicit separation of abstract vs concrete here.

---

This PR is best reviewed commit-by-commit. :)

### DIFF
--- a/src/resolvelib/__init__.py
+++ b/src/resolvelib/__init__.py
@@ -16,8 +16,8 @@ __version__ = "1.0.2.dev0"
 
 from .providers import AbstractProvider
 from .reporters import BaseReporter
-from .resolvers import (
-    AbstractResolver,
+from .resolvers.abstract import AbstractResolver
+from .resolvers.criterion import (
     InconsistentCandidate,
     RequirementsConflicted,
     ResolutionError,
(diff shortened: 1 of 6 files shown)


src/resolvelib/__init__.py
```diff
@@ -16,8 +16,8 @@
 
 from .providers import AbstractProvider
 from .reporters import BaseReporter
-from .resolvers import (
-    AbstractResolver,
+from .resolvers.abstract import AbstractResolver
+from .resolvers.criterion import (
```
frostming, line 20: These should be imported from `.resolvers`, where all required members should be exported.

src/resolvelib/resolvers/criterion.py
```diff
@@ -27,17 +19,10 @@
     State,
     build_iter_view,
 )
+from .abstract import AbstractResolver, Result
```
frostming, line 22: The implementations(Resolution/Resolver) should belong to a separate module, with the name like `resolvers/impl.py`

PR #152, opened 2024-06-14
Fix backjumping by checking if broken_state.mapping is empty

notatallshaw:

Still need to provide a test, going to see if I can write a small provider that can process the JSON in https://github.com/sarugaku/resolvelib/issues/134#issuecomment-2164347177 and follows pip's preference logic. 

But feedback would be welcome.

### DIFF
--- a/src/resolvelib/resolvers.py
+++ b/src/resolvelib/resolvers.py
@@ -309,23 +309,28 @@ class Resolution(Generic[RT, CT, KT]):
             # Remove the state that triggered backtracking.
             del self._states[-1]
 
-            # Ensure to backtrack to a state that caused the incompatibility
-            incompatible_state = False
+            # Optimistically backtrack to a state that caused the incompatibility
             broken_state = self.state
-            while not incompatible_state:
+            while True:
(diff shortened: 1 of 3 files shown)


notatallshaw:
Upon writing a test case I realized it wasn't sufficent to check if name is in incomatible dependencies, but the simpler check of checking if broken_state.mapping is empty is sufficent.

notatallshaw:
Actually, while this solves the given test case, I found it still fails for "kedro[test]==0.18.13".

I am going to leave this PR as ready to review, because I still beleive it's correct to exist from backtracking when broken_state.mapping is empty.

But I am going to try and create another minimal example from "kedro[test]==0.18.13".

notatallshaw:
> I found it still fails for "kedro[test]==0.18.13"

After a bit of investigation I found that this is a weird edge case where the extra itself has contradictory requirement, and the bug might be pip specific and how it passes extra constraints to resolvelib.

I think that should be treated as a completely seperate issue, that I will file a bug for on pip side and investigate.

uranusjr (approved):

PR #154, opened 2024-06-17
Backjumping failures

notatallshaw:
Here are two tests (I may add more) that fail because of backjumping.

None of these tests are fixed by https://github.com/sarugaku/resolvelib/pull/152 because that is only able to recover the situation if each requirement has remaining unchecked candidates, but backjumping can backtrack to the root requirement, incorrectly, and if, for example, that requirement has only one candidate then there can be an erroneous ResolutionImpossible.

I *believe* the issue is that backjumping sets up an implicit contract, caused by [this](https://github.com/sarugaku/resolvelib/blob/main/src/resolvelib/resolvers.py#L326), between itself and the provider. That the provider should always prefer direct conflicts over any other candidate. However, resolvelib does not enforce or document this, and the provider can prefer any unsatisfied name, and therefore the algorithm is currently unsound, for example pip does not prefer directly conflicting packages, at least not [yet](https://github.com/pypa/pip/pull/12499).

@uranusjr @frostming @bennati IMO either backjumping should be reverted, or it should be made opt-in for providers who "know what they are doing".

### DIFF
--- a/tests/functional/python/inputs/case/issue-134-2.json
+++ b/tests/functional/python/inputs/case/issue-134-2.json
@@ -1,0 +1,15 @@
+{
+	"index": "issue-134-2",
+	"requested": [
+		"parent"
+	],
+	"resolved": {
+		"parent": "0.0.1",
+		"pandas": "1.3.5",
+		"pystac": "1.8.3",
+		"pystac-client": "0.3.3",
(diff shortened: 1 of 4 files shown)


notatallshaw:
I have put these test cases into: https://github.com/sarugaku/resolvelib/pull/155

PR #157, opened 2024-07-30
Replace setuptools.cfg and mypy.ini with pyproject.toml

notatallshaw:
It's probably time to migrate everything over to pyproject.toml.

After this I plan to look at updating the pinned version of black and/or switching to ruff for formating and/or linting.


pyproject.toml
```diff
@@ -1,5 +1,63 @@
...
+include-package-data = true
+
+[tool.setuptools.packages]
+find = {where = ["src"], namespaces = false}
+
+[tool.setuptools.dynamic]
+version = {attr = "resolvelib.__version__"}
+
+[tool.distutils.bdist_wheel]
+universal = true
```
webknjaz, line 59: With the project only supporting Python 3 now, this should be false. Otherwise, the wheels will keep being tagged as `py2.py3`.

webknjaz: https://github.com/sarugaku/resolvelib/pull/178

pyproject.toml
```diff
@@ -1,5 +1,63 @@
...
+    "pytest",
+]
+release = [
+    "build",
+    "towncrier",
+    "twine",
+]
+
 [build-system]
 requires = ['setuptools>=62', 'wheel']
```
webknjaz, line 46: https://github.com/sarugaku/resolvelib/pull/175

PR #166, opened 2024-08-01
Guard type hinting via ruff

notatallshaw:
This PR builds on top of https://github.com/sarugaku/resolvelib/pull/165

I notice it's a common pattern in this library, and one I prefer, to use `from __future__ import annotations` and guard type hint only imports behind `if TYPE_CHECKING`, this can be implemented automatically with ruff using: `required-imports = ["from __future__ import annotations"]` with the isort rules and the TCH rules: https://docs.astral.sh/ruff/rules/#flake8-type-checking-tch (namely TCH001, TCH002, and TCH003).

I updated the config than then just ran a series of ruff --fix commands for the remaining commits (I have documented them to be clear).

Let me know if this is to your liking or not.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -84,7 +84,7 @@ src = ["src"]
 line-length = 88
 
 [tool.ruff.lint]
-select = ["C","E","F","W","B","RUF","PLE","PLW","I"]
+select = ["C","E","F","W","B","RUF","PLE","PLW","I","TCH","FA"]
 ignore = ["PLW2901"]
 exclude = [
 	".git",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -80,10 +80,11 @@ name = 'Trivial Changes'
 showcontent = false
 
 [tool.ruff]
+src = ["src"]
 line-length = 88
 
 [tool.ruff.lint]
-select = ["C","E","F","W","B","RUF","PLE","PLW"]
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -97,7 +98,7 @@ exclude = [
 ]
 
 [tool.ruff.lint.isort]
-known-first-party = ["resolvelib"]
+required-imports = ["from __future__ import annotations"]
```
edgarrmondragon, line 101: There's also the [`FA` rules](https://docs.astral.sh/ruff/rules/#flake8-future-annotations-fa), which might make this setting redundant :)

notatallshaw: In the past I found FA rules didn't always enable `from __future__ import annotations` when I expected, but looking at this PR it seems like all the additional `from __future__ import annotations`  aren't stictly necessary.

I've updated to use FA instead, it makes the PR a lot less noisey. Thanks.

notatallshaw:
FYI, this actually appears to catch a bug by correctly moving `RequirementsConflicted` out of the `TYPE_CHECKING` block in `tests/test_resolvers.py`, as it used in an exception catch.

It seems like not test case is triggering this test code path, if no one else does I will take a look in the coming weeks.

pradyunsg (approved):

Issue #159, opened 2024-08-01
Request: New Release


notatallshaw:
Hi all, given the recent fixes to backjumping I am requesting a new release. I would like to vendor resolvelib early into pip comfortably before 24.3 so it can be sufficently tested.

I also notice it's been over a year since release, so there are quite a few PRs that have been merged:

e4de27e - Merge pull request #158 from notatallshaw/add-unvisited-tests
58c5d90 - Merge pull request #155 from notatallshaw/skip-backjumping-when-name-is-not-incompatible
314dde4 - Merge pull request #157 from notatallshaw/replace-setuptools-cfg-with-pyproject.toml
b01ff4e - Merge pull request #156 from notatallshaw/fix-up-py2index-script
ffa7199 - Merge pull request #152 from notatallshaw/fix-backjumping
adf0800 - Merge pull request #153 from notatallshaw/run-python-3.12-and-3.13-and-bump-all-versions-in-ci.yml
87c9d7b - Merge pull request #141 from hroncok/nocmmments
a062162 - Merge pull request #150 from notatallshaw/pin-black-version
995ed67 - Merge pull request #143 from notatallshaw/deduplicate-causes
044ab9f - Merge pull request #135 from sarugaku/modern-typing
6df10d7 - Merge pull request #136 from jimkring/patch-1
69d9fee - Merge pull request #138 from brettcannon/clarify-docstrings
77b256c - Merge pull request #133 from hauntsaninja/hauntsaninja-patch-1
ea4011b - Merge pull request #129 from OlenaYefymenko/122-supported-python-versions

frostming:
@notatallshaw Thank you very much. Also, if you have time, could you add the missing news fragments? They should be located under the news/ folder

notatallshaw:
> @notatallshaw Thank you very much. Also, if you have time, could you add the missing news fragments? They should be located under the news/ folder

I never understood how news fragments worked, I'll do a bit of reading this evening and hopefully be able to add them.

frostming:
It seems all issues are addressed. But I don't have the permission to upload releases, nor is the token configured in the GitHub secrets. I'll leave it to @pradyunsg and @uranusjr to make a new release.

notatallshaw:
Thanks for your assistance! 

Maybe we should expand the number of maintainers with access to upload? and also I would be happy to help maintain the project., if they would help avoid the project accidentally running out of people with access to release.

pradyunsg:
[...]

PR #168, opened 2024-08-14
chore(ci): setup automated release workflow

frostming:
Signed-off-by: Frost Ming <me@frostming.com>

Related to #159

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,35 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - "*"
+
+jobs:
+  build-package:
+    name: Build & inspect our package.
+    runs-on: ubuntu-latest
+
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,35 @@
...
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - uses: hynek/build-and-inspect-python-package@v2
+
+  release-pypi:
+    name: Publish released package to pypi.org
+    permissions:
+      id-token: write
+    environment: release
```
webknjaz, line 23: FTR, I recommend calling the environment `pypi` since it represents a deployment target which is PyPI. Also, it's nice to also show a URL on the UI.

webknjaz: https://github.com/sarugaku/resolvelib/pull/177

Issue #171, opened 2024-10-24
Causes returned by `_attempt_to_pin_criterion` are too broad (causing searching unnecessary parts of a dependency graph)


notatallshaw:
In a complex resolution the causes returned by `_attempt_to_pin_criterion` when rejecting a candidate are far too broad, if we take a look at the example in https://github.com/pypa/pip/issues/13037#issue-2605712321 I've created a log that prints out each rejection: https://github.com/pypa/pip/issues/13037#issuecomment-2433783030

Looking at one of the rejections:

```
Rejecting thinc 8.3.1, due to conflict:
	The user requested numpy==1.21.5
	spacy 3.8.2 depends on numpy>=1.19.0; python_version >= "3.9"
	mlflow 2.17.0 depends on numpy<3
	matplotlib 3.8.4 depends on numpy>=1.21
	pandas 2.0.3 depends on numpy>=1.21.0; python_version >= "3.10"
	pyarrow 17.0.0 depends on numpy>=1.16.6
	scikit-learn 1.5.2 depends on numpy>=1.19.5
	scipy 1.10.1 depends on numpy<1.27.0 and >=1.19.5
	thinc 8.3.1 depends on numpy<2.1.0 and >=2.0.0; python_version >= "3.9"
```

This should be ideally narrowed to:
```
	The user requested numpy==1.21.5
	thinc 8.3.1 depends on numpy<2.1.0 and >=2.0.0; python_version >= "3.9"
```

This would allow downstream libraries (like pip) that use cases to prefer backtracking to hone in on the correct causes, and it would produce a much more focused message to the user, especially when hitting an impossible resolution.

Because resolvelib is so generic a cause narrowing is a little tricky, but I propose the following steps:

1. While there are more than 2 causes loop through each of the causes
2. Remove that cause and confirm there is still a conflict
3. Check there is no conflict between this cause and any other cause
4. If the above criteria pass remove it from the cause list

I think this could be a huge speed up for very problamatic resolutions where the downstream library uses the causes to determine what to prefer, I will look to make a PR in the coming weeks.

notatallshaw:
FYI, I'm not sure in resolvelib's current design cause narrowing like this is possible, but if it can happen at the resolvelib level I think it would be best as then it benefits all clients and keeps less of the resolution optimization logic out of the downstream library. So I've opened this issue somewhat optimistically, and I'll report back on my findings once I try.

frostming:
[...]

PR #179, opened 2024-11-10
New disjoint method for provider

notatallshaw:
Fixes https://github.com/sarugaku/resolvelib/issues/171

This allows the provider to implement a method to determine the causes which are disjoint, this has a number of advantages:

1. It reduces the number of causes saved to the state
2. If the provider uses `backtrack_causes` in `get_preference` or `narrow_requirement_selection` they will be the disjoint causes allowing the provider to not have to iterate through as many causes or re-implement figuring out disjoint causes on every different method call
3. For `ResolutionImpossible` errors the disjoint causes will naturally be returned 

If the provider does not implement a `disjoint` method then there is no behavior change, so this PR should be completely backwards compatible.

### DIFF
--- a/src/resolvelib/providers.py
+++ b/src/resolvelib/providers.py
@@ -194,3 +194,30 @@ class AbstractProvider(Generic[RT, CT, KT]):
             Iterable[KT]: A non-empty subset of `identifiers`.
         """
         return identifiers
+
+    def disjoint(
+        self, backtrack_causes: Sequence[RequirementInformation[RT, CT]]
+    ) -> Sequence[RequirementInformation[RT, CT]]:
+        """Filter backtrack causes to retain only disjoint instances.
+
+        Given a sequence of backtrack causes, return those that are disjoint.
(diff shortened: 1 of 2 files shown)


notatallshaw:
This new method seems quite natrual, happy to take any bikeshedding suggestions.

I have a hacky version of implementing it with pip and dep-logic: https://github.com/pypa/pip/compare/main...notatallshaw:pip:speedy-resolve (For the real version I will need to more carefully consider when ExplicitRequirements and SpecifierRequirements are disjoint). But it already successfully speeds up various resolutions.

notatallshaw:
I've come to realize a problem with this approach, or at least why pip couldn't easily adopt it, I write up in more detail shortly.

notatallshaw:
I no longer think this is the right approach, this is trying to patch the problem after the fact, it would be better to implement something when working out the causes initially, e.g. https://github.com/sarugaku/resolvelib/issues/180#issuecomment-2573136409

PR #181, opened 2024-12-02
Fix pypi project page link on README.rst

dholth:
This should fix the link on https://pypi.org/project/resolvelib/

Is there a better way to fix this link?


notatallshaw:
This seems reasonable to me, I notice that all of pip's links are absolute: https://raw.githubusercontent.com/pypa/pip/refs/heads/main/README.rst

I've fixed the Ruff linting error, can you please rebase or merge with main.

Issue #180, opened 2024-12-02
Back jump doesn't do well in some cases


frostming:
Here are the package set to be resolved

```
1. A==1
2. B  # B@2 depends on D@2 and B@1 depends on D@1
3. C
4. D  # D@2 depends on A@2 and D@1 depends on A@1
```
Assume the resolver resolves packages from top to down, when resolving D, the current pins are:

```
A@1
B@2
C@x
```

Because `B@2` is pinned which constrains `D==2`, we can't find a compatible version of `D` that is also happy with `A@1`, so conflicts happen and we back jump and pop the last pinned package `C`, which is totally unrelated to the incompatibilities.

However, due to the line in `backjump` implementation: https://github.com/sarugaku/resolvelib/blob/51dbc929523577aea960dd3672df49c9a428ae7d/src/resolvelib/resolvers/resolution.py#L327-L330

The backjump quits very early and fallbacks to backtrack. It would be extremely slow if traversing C is super expensive. But it's obvious we should jump over to the state of `B` and try other versions. My question is, does the quit condition make sense in any case I am not aware of or should we fix it?

@notatallshaw 

- - -

To be concrete, `A: h11, B: httpcore, C: Pillow, D: httpx`

notatallshaw:
Yeah, the exit is probably too aggressive, but there are definitely cases that incorrectly reach resolution impossible without it.

I have at least a few real world examples to try against if you have an idea how to loosen this without causing incorrect resolutions.

frostming:
> I have at least a few real world examples to try against if you have an idea how to loosen this without causing incorrect resolutions.

I think we previously proposed to save the states first, and if no state that can jump is found, we restore to backtrack.

notatallshaw:
Btw, I have an _intuition_ that if the backtrack causes are disjoint, i.e every cause is incomparable with some other cause (or set of causes), resolvelib could aggressively backjump. But currently the provider has no way to indicate to the resolver disjointness.

I will experiment this week with disjointness and aggressively backjumping, to see if the intuition is correct.

notatallshaw:
> I think we previously proposed to save the states first, and if no state that can jump is found, restore the states and fallback to backtrack.

Yeah, it wasn't very elegant though, for example it made complicated impossible resolutions much slower.

notatallshaw:
[...]

PR #183, opened 2025-01-07
Optimistic Backjumping

notatallshaw:
Fixes https://github.com/sarugaku/resolvelib/issues/180

Implements an optimistic backjumping strategy where it is assumed to be logically sound unless a Resolution Impossible is hit, then reverts back to the state where optimistic backjumping first diverged from regular backjumping/backtracking, and tries again. This speeds up a lot of scenarios, but isn't a 100% win.

I would like to add tests similiar to https://github.com/sarugaku/resolvelib/pull/158 as not to regress with any future improvements, I will see if I can figure out some.


notatallshaw:
Closing in favor of https://github.com/sarugaku/resolvelib/pull/188

PR #188, opened 2025-05-28
Optimistic backjumping with cutoff

notatallshaw:
Fixes https://github.com/sarugaku/resolvelib/issues/180

This implements https://github.com/sarugaku/resolvelib/issues/180#issuecomment-2847584328 with a ratio of 10%.

10% is somewhat arbitrary, but the general idea is optimistic backtracking is supposed to be much faster than regular backtracking, and therefore if it is taking a significant amount of the remaining rounds left it is *probably* going down the wrong path. I tried several different numbers for different max rounds on real world and synthetic cases and found between 5% and 20% tended to work best.

While not intended to be changed I have made `_OPTIMISTIC_BACKJUMPING_RATIO` a global as an easy way for a knowledgeable motivated resolvelib user to pick a different ratio (in particular they may want to pick 0 or 1).

I have added additional tests that fail without optimistic backjumping (i.e. optimistic backjumping is able to skip over certain requirements that regular backjumping can not). I also test turning off optimistic backjumping to make sure all other tests pass without it.

(Note: I have collected recent problematic real world examples in https://github.com/pypa/pip/issues/13281 and have been documenting them in [problematic.toml](https://github.com/notatallshaw/Pip-Resolution-Scenarios-and-Benchmarks/blob/main/scenarios/problematic.toml)).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -98,7 +98,7 @@ exclude = [
 ]
 
 [tool.ruff.lint.mccabe]
-max-complexity = 12
+max-complexity = 20
 
 [tool.mypy]
 warn_unused_configs = true
--- a/src/resolvelib/resolvers/resolution.py
+++ b/src/resolvelib/resolvers/resolution.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import collections
 import itertools
 import operator
-from typing import TYPE_CHECKING, Collection, Generic, Iterable, Mapping
+from typing import TYPE_CHECKING, Generic
 
 from ..structs import (
     CT,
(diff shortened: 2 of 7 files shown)


notatallshaw:
@uranusjr @frostming Any thoughts on this?

In an ideal world I would like to see this released so I can get this in the next version of Python. It solves 7 out of the 8 ResolutionTooDeep errors I have open against pip without introducing any new ones.

frostming:
LGTM from my side.

notatallshaw:
Okay, I'll leave it a few more days if anyone else wants to comment and then merge.

And then look to do a release shortly there after (not sure if I have required permissions, will open an issue for a release and we'll see how it goes).

PR #193, opened 2025-10-01
Change `Mapping` to `dict` in resolution `Result` type annotations

webknjaz:
`typing.Mapping` is a broad concept. The actual code just assigns `dict`s so `Result` should just use that.

I've been adding annotations to the calling side (`ansible-galaxy`) that uses `resolvelib` and hit this as I was trying to show that a function that returns `Resolver(...).resolve(...).mapping` produces a `dict`.

### DIFF
--- a/src/resolvelib/resolvers/abstract.py
+++ b/src/resolvelib/resolvers/abstract.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import collections
-from typing import TYPE_CHECKING, Any, Generic, Iterable, Mapping, NamedTuple
+from typing import TYPE_CHECKING, Any, Generic, Iterable, NamedTuple
 
 from ..structs import CT, KT, RT, DirectedGraph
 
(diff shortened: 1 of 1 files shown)


notatallshaw:
Is it not expected that anyone is sub-classing `AbstractResolver` to implement their own `Resolver`? I don't think pip is, but if anyone else is this technically changes the API for resolvelib.

Could this be done at the `Resolver(...).resolve(...) level instead? Or does that break some type theory invariant?

I'm not against this, just thinking about where things could go wrong.

webknjaz:
@notatallshaw AFAIK, only the `AbstractProvider` needs to be inherited. Also, if somebody is substituting entire main components of a library they are effectively not using the library in the end.

This cannot be done on the `Resolver(...).resolve(...)` because it returns a `Result` object which has a broadly typed attribute. Currently, I'm fighting this with type narrowing: https://github.com/ansible/ansible/pull/85545/files#r2394287530. But I think the types should be fixed here. After all, other components in the repo use `dict`. So this seems inconsistent.

frostming:
The broad type is intentional according to [SOLID](https://en.wikipedia.org/wiki/SOLID) principles, specifically:

>no code should be forced to depend on [methods](https://en.wikipedia.org/wiki/Method_(computer_programming)) it does not use.[[1]](https://en.wikipedia.org/wiki/Interface_segregation_principle#cite_note-ASD-1) ISP splits [interfaces](https://en.wikipedia.org/wiki/Interface_(computing)) that are very large into smaller and more specific ones so that clients will only have to know about the methods that are of interest to them.

You can easily turn it into a dict using `dict(result)` in the return statement, though with a little overhead as it creates a copy.

PR #204, opened 2026-03-18
Skip orphaned criteria in unsatisfied_names

notatallshaw:
When `_remove_information_from_criteria` empties a criterion's information list, skip it in `unsatisfied_names` instead of pinning it and its transitive deps for nothing.

Found via pip on `kedro[test]==0.18.13` (12 orphaned packages pinned then discarded by `_build_result`).


notatallshaw:
@frostming I found this while investigating performance optimizations, it usually requires fairly complex backtracking scenarios to trigger in my testing (barely triggers in many of the backtracking scenarios I have for pip).

